PR #8, opened 2020-03-24
Give build stop functionality using CLI

nagpalkaran95:


bin/runner.js
```diff
@@ -37,9 +37,9 @@ var argv = yargs
       return require('./commands/init')(argv);
     }
   })
-  .command('build', Constants.cliMessages.BUILD.INFO, function(yargs) {
+  .command('build-info', Constants.cliMessages.BUILD.INFO, function(yargs) {
```
SagarGaniga: are we changing this command name? We will have to update the documentation as well for this

nagpalkaran95: done

PR #13, opened 2020-04-22
[Issue #11] Support for localIdentifier capability

### LINKED ISSUE #11: localIdentifier capability should be supported with the local capability
since the BrowserStack Local capability is supported in the connection_settings configuration, you should also provide the ability to set a value for *localIdentifier* as this would be more in line with how the Cypress tests would need to be used if they're part of a CICD pipeline that runs multiple builds in parallel.
Example implementation should look like the following:
```json
{
   "connection_settings": {
      "local": true,
      "localIdentifier": "build_name+build_number"
    }
}
```
this would then allow for adding a reference to the browserstack-local npm package which could be started using the same localIdentifier and accessKey as are used for the Cypress tests

nagpalkaran95:
Description: Provides support for localIdentifier capability to be used while running local sessions

Issue Link: https://github.com/browserstack/browserstack-cypress-cli/issues/11

- Fix #11


bin/helpers/capabilityHelper.js
```diff
@@ -33,6 +33,11 @@ const caps = (bsConfig, zip) => {
     if (bsConfig.connection_settings.local === true) obj.local = true;
     logger.log(`Local is set to: ${obj.local}`);
 
+    // Local Identifier
+    obj.localIdentifier = null;
+    if (bsConfig.connection_settings.localIdentifier) obj.localIdentifier = bsConfig.connection_settings.localIdentifier;
```
suryart, line 38: Check local is true as well.

README.md
```diff
@@ -80,6 +81,7 @@ The following table provides a reference for all the options that can be provide
...
 | `local`           | Helps in testing websites that cannot be accessed in public network. If you set this to `true`, please download the Local binary and establish a local connection first (you can learn how to do so [here](https://www.browserstack.com/local-testing/automate#command-line))                                                                       | Boolean: `true` / `false`. Set this to `true` if you need to test a local website. Set this to `false` if the website is accessible publicly.              |
+| `localIdentifier`           |  Configure the test script to run through a specific connection. To do so, set the value of localIdentifier capability to be a unique connection name.                                                                       | Set this when `local` is set to `true`.              |
```
srebs, line 84: Change the text in 2nd column to:

The BrowserStack Local tunnel that should be used to resolve requests. This is applicable only when you start a Local binary with a local identifier. Your tests might fail if you use an invalid local identifier. This option will be ignored if `local` option is set to `false`.


Change the text in 3rd column to:

A string if a tunnel identified by the Local identifier should be used (should be same as the one used to start the Local binary). Set this to `null` (default value) to resolve requests using Local without a Local identifier.

PR #14, opened 2020-05-05
Adding required instrumentation

nagpalkaran95:
**User facing changelog**
* Colorised logging
    * info will be green
    * error will be red

**PR Tasks**
* [x] Collect anonymized usage data including the command-line arguments
used, system details and errors that users get so that we can improve the way
users run their Cypress tests on BrowserStack.
* [x] Update package imports as per nodejs best practices
* [x] Migrate to winston logger from custom logger
* [x] Handle case when `browserstack.json` is an invalid json

### DIFF
--- a/bin/commands/info.js
+++ b/bin/commands/info.js
@@ -20,12 +20,15 @@ function buildInfo(args) {
 
     let options = {
       url: config.buildUrl + buildId,
-      method: 'GET',
+      method: "GET",
       auth: {
         user: bsConfig.auth.username,
-        password: bsConfig.auth.access_key
-      }
-    }
+        password: bsConfig.auth.access_key,
+      },
+      headers: {
+        "User-Agent": util.getUserAgent,
+      },
+    };
 
     request(options, function (err, resp, body) {
       let message = null;
--- a/bin/commands/info.js
+++ b/bin/commands/info.js
@@ -46,21 +46,9 @@ function buildInfo(args) {
           build = null;
         }
 
-        if (resp.statusCode != 200) {
-          messageType = Constants.messageTypes.ERROR;
-          errorCode = 'api_failed_build_info';
-
-          if (build) {
-            message = `${Constants.userMessages.BUILD_INFO_FAILED} with error: \n${JSON.stringify(build, null, 2)}`;
-            logger.error(message);
-            if (build.message === 'Unauthorized') errorCode = 'api_auth_failed';
-          } else {
-            message = Constants.userMessages.BUILD_INFO_FAILED;
-            logger.error(message);
-          }
-        } else if (resp.statusCode == 299) {
+        if (resp.statusCode == 299) {
           messageType = Constants.messageTypes.INFO;
-          errorCode = 'api_deprecated';
+          errorCode = "api_deprecated";
 
           if (build) {
             message = build.message;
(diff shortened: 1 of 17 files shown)


bin/helpers/usageReporting.js
```diff
@@ -0,0 +1,209 @@
...
+  // Visual Studio Team Services
+  if (env.TF_BUILD === "True") {
+    return "Visual Studio Team Services";
+  }
+  // if no matches, return null
+  return null;
+}
+
+function isUsageReportingEnabled() {
+  return process.env.DISABLE_USAGE_REPORTING;
```
suryart, line 155: Let's document how to opt-out of the usage reporting, please.

Also, it's worth mentioning what all will be collected, and everything is anonymous and cannot be used for identifying a run.

bin/helpers/usageReporting.js
```diff
@@ -0,0 +1,209 @@
...
+    }
+    return [version, _path];
+  } else {
+    let version = get_version('browserstack-cypress');
+
+    if (!version) {
+      // return path = null if version is null
+      return [null, null];
+    }
+    return [version, npm_global_path()];
```
suryart, line 104: Instead of an array, return an object.

bin/helpers/logger.js
```diff
@@ -1,12 +1,35 @@
-function log(message) {
-  var timestamp = '[' + new Date().toLocaleString() + '] ';
-  console.log(timestamp + " " + message);
-}
+const winston = require('winston'),
+  fs = require("fs"),
+  path = require("path");
+
+const { format } = winston;
+// const { format } = require("logform");
```
suryart, line 6: remove commented code if no longer needed.

bin/helpers/config.js
```diff
@@ -11,5 +11,7 @@ config.rails_host = hosts[config.env].rails_host;
 config.cypress_v1 = `${config.rails_host}/automate/cypress/v1`;
 config.buildUrl = `${config.cypress_v1}/builds/`;
 config.buildStopUrl = `${config.cypress_v1}/builds/stop/`;
+config.usageReportingUrl = `http://127.0.0.1:8000/send_event_cy_internal`;
```
suryart, line 14: What http://127.0.0.1:8000 points to?

bin/commands/info.js
```diff
@@ -1,55 +1,85 @@
...
+
+          if (build) {
+            message = `${Constants.userMessages.BUILD_INFO_FAILED} with error: \n${JSON.stringify(build, null, 2)}`;
+            logger.error(message);
+            if (build.message === 'Unauthorized') errorCode = 'api_auth_failed';
+          } else {
+            message = Constants.userMessages.BUILD_INFO_FAILED;
+            logger.error(message);
+          }
+        } else if (resp.statusCode == 299) {
```
SagarGaniga, line 61: This condition is not reachable. We should move it up

bin/helpers/config.js
nagpalkaran95: Will point to the endpoint where data will be sent. Will update this before merging

bin/helpers/usageReporting.js
nagpalkaran95: done

bin/helpers/util.js
```diff
@@ -0,0 +1,75 @@
...
+exports.setUsageReportingFlag = (bsConfig, disableUsageReporting) => {
+  if (disableUsageReporting === undefined && bsConfig && bsConfig['disable_usage_reporting'] != undefined) {
+    process.env.DISABLE_USAGE_REPORTING = bsConfig['disable_usage_reporting'];
+  } else {
+    process.env.DISABLE_USAGE_REPORTING = disableUsageReporting;
+  }
+}
+
+exports.getUserAgent = () => {
+  return `BStack-Cypress-CLI/1.x (${os.arch()}/${os.platform()}/${os.release()})`;
```
srebs, line 74: @nagpalkaran95 Can we get the exact version of the CLI? 

Like `BStack-Cypress-CLI/<full_version> (${os.arch()}/${os.platform()}/${os.release()})`?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "browserstack-cypress-cli",
-  "version": "1.1.0",
+  "version": "1.1.3",
```
suryart, line 3: This is handled during the package publishing, not needed.

PR #15, opened 2020-05-08
Installing npm packages support win10

suryart:
Since mostly Cypress tests are written with external npm packages. We would like to enable the CLI user to be able to let us know the dependencies so that we can install them. Currently, this is limited only for Windows 10 machines.

This PR includes:

- Adds `npm_depedencies` key to the browserstack.json config file.


bin/helpers/archiver.js
```diff
@@ -35,9 +35,13 @@ const archiveSpecs = (runSettings, filePath) => {
...
     archive.pipe(output);
 
+    var packageJSON = JSON.stringify({devDependencies: bsConfig.npm_dependencies}, null, 4);
+
     let allowedFileTypes = [ 'js', 'json', 'txt', 'ts' ]
     allowedFileTypes.forEach(fileType => {
-      archive.glob(`**/*.${fileType}`, { cwd: cypressFolderPath, matchBase: true, ignore: 'node_modules/**' });
+      archive
+        .glob(`**/*.${fileType}`, { cwd: cypressFolderPath, matchBase: true, ignore: 'node_modules/**' })
+        .append(packageJSON, { name: 'package.json' });
```
SagarGaniga, line 44: can we take this out of forEach?

bin/helpers/archiver.js
```diff
@@ -36,10 +36,13 @@ const archiveSpecs = (runSettings, filePath) => {
 
     archive.pipe(output);
 
+    var packageJSON = JSON.stringify({devDependencies: runSettings.npm_dependencies}, null, 4);
```
SagarGaniga, line 39: need to handle the case where the user is not setting any `npm_dependencies` or just removed the key manually.

suryart: Done

PR #25, opened 2020-05-31
[CYP-105] Unit tests

nagpalkaran95:
Adds unit tests with 80% code coverage


bin/commands/info.js
```diff
@@ -4,33 +4,28 @@ const request = require('request');
 const config = require("../helpers/config"),
   logger = require("../helpers/logger").winstonLogger,
   Constants = require("../helpers/constants"),
-  util = require("../helpers/util");
+  utils = require("../helpers/utils");
```
suryart, line 7: Why the pluralization?

nagpalkaran95: Because `util` is inbuilt node.js module. Hence changed it to `utils` for user defined module.

bin/helpers/utils.js
```diff
@@ -53,6 +53,7 @@ exports.getErrorCodeFromErr = (err) => {
 }
 
 exports.sendUsageReport = (bsConfig, args, message, message_type, error_code) => {
+  console.log("okay")
```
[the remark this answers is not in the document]
nagpalkaran95: removed

PR #28, opened 2020-06-12
Adding validation to check if cypress.json is present at the cypress_proj_dir.

karanshah81295:
Adding validation to check if cypress.json is present at the cypress_proj_dir.

### DIFF
--- a/bin/helpers/capabilityHelper.js
+++ b/bin/helpers/capabilityHelper.js
@@ -93,7 +93,7 @@ const validate = (bsConfig) => {
 
     if(!bsConfig.run_settings.cypress_proj_dir) reject(Constants.validationMessages.EMPTY_SPEC_FILES);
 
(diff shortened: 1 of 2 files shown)


bin/helpers/capabilityHelper.js
```diff
@@ -91,6 +93,8 @@ const validate = (bsConfig) => {
 
     if(!bsConfig.run_settings.cypress_proj_dir) reject(Constants.validationMessages.EMPTY_SPEC_FILES);
 
+    if (!fs.existsSync(path.join(process.cwd(), bsConfig.run_settings.cypress_proj_dir, 'cypress.json'))) reject(Constants.validationMessages.EMPTY_CYPRESS_JSON);
```
suryart, line 96: Can we also include the path as well in the error message `bsConfig.run_settings.cypress_proj_dir`?

karanshah81295: Have also added the Cypress Project Directory in the error message.

PR #29, opened 2020-06-16
Cyp 175 parallelisation support

pawanpanjwani:
Parallelization support for cypress


bin/helpers/capabilityHelper.js
```diff
@@ -91,7 +95,13 @@ const validate = (bsConfig) => {
 
     if (!bsConfig.run_settings) reject(Constants.validationMessages.EMPTY_RUN_SETTINGS);
 
-    if(!bsConfig.run_settings.cypress_proj_dir) reject(Constants.validationMessages.EMPTY_SPEC_FILES);
+    if (!bsConfig.run_settings.cypress_proj_dir) reject(Constants.validationMessages.EMPTY_SPEC_FILES);
+
+    // validate parallels specified in browserstack.json if parallels are not specified via arguments
+    if (!Utils.isUndefined(args) && Utils.isUndefined(args.parallels) && !Utils.isParallelValid(bsConfig.run_settings.parallels)) reject(Constants.validationMessages.INVALID_PARALLELS_CONFIGURATION);
```
nagpalkaran95, line 101: do we really need `!Utils.isUndefined(args)` check?

bin/helpers/capabilityHelper.js
```diff
@@ -91,7 +95,13 @@ const validate = (bsConfig) => {
...
+  
+    // if parallels specified via arguments validate only arguments
+    if (!Utils.isUndefined(args) && !Utils.isUndefined(args.parallels) && !Utils.isParallelValid(args.parallels)) reject(Constants.validationMessages.INVALID_PARALLELS_CONFIGURATION);
```
nagpalkaran95, line 104: same as comment on line 101

bin/helpers/capabilityHelper.js
pawanpanjwani: if`args` would be null or undefined , in that case we would be trying to access parallels of null/undefined which would eventually break

pawanpanjwani: Adding a safe-check just for the case

PR #31, opened 2020-07-01
[CYP-236] Change console message string if testings mins are expired

nagpalkaran95:
**User facing changelog**
*  Update the error message from `internal server error` to `Automate testing minutes expired.` if automate testing mins are expired for the user


nagpalkaran95:
Not required now. Handled at back end.

PR #37, opened 2020-07-10
Add link to dashboard on build creation

nagpalkaran95:
### User facing changelog
* Add a log line with dashboard link when build is created successfully

### DIFF
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -38,9 +38,12 @@ module.exports = function run(args) {
         return zipUploader.zipUpload(bsConfig, config.fileName).then(function (zip) {
 
           // Create build
-          return build.createBuild(bsConfig, zip).then(function (message) {
+          return build.createBuild(bsConfig, zip).then(function (data) {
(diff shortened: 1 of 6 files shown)


bin/commands/runs.js
```diff
@@ -38,9 +38,12 @@ module.exports = function run(args) {
         return zipUploader.zipUpload(bsConfig, config.fileName).then(function (zip) {
 
           // Create build
-          return build.createBuild(bsConfig, zip).then(function (message) {
+          return build.createBuild(bsConfig, zip).then(function (data) {
+            let message = `${data.message}! ${Constants.userMessages.BUILD_CREATED} with build id: ${data.build_id}`;
+            let dashboardLink = `Visit the Automate dashboard for test reporting: ${config.dashboardUrl}${data.build_id}`;
```
SagarGaniga, line 43: `Visit the Automate dashboard for test reporting`

Can we put this into constants as this might change in future and can be used in other commands as well, for example in stop also we can show the URL to that stopped build.

PR #43, opened 2020-07-21
Make config management developer friendly

nagpalkaran95:
### PR Tasks
* Changing the way we load config to make it more development friendly

### DIFF
--- a/bin/helpers/config.js
+++ b/bin/helpers/config.js
@@ -1,16 +1,25 @@
 var config = {};
 config.env = process.env.NODE_ENV || "production";
 
-// load config based on env
-require('custom-env').env(config.env)
+if(config.env === "production") {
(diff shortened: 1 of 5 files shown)


bin/helpers/config.js
```diff
@@ -1,16 +1,25 @@
 var config = {};
 config.env = process.env.NODE_ENV || "production";
 
-// load config based on env
-require('custom-env').env(config.env)
+if(config.env === "production") {
+  config.uploadUrl = 'https://api-cloud.browserstack.com/automate-frameworks/cypress/upload';
```
suryart, line 5: Can we move these to config.json and load that JSON on line 1?

nagpalkaran95: done

PR #48, opened 2020-08-26
Cyp 372 env var

karanshah-browserstack:
Exporting BUILD_ID & BUILD_URL to build_results.txt

### DIFF
--- a/bin/helpers/utils.js
+++ b/bin/helpers/utils.js
@@ -131,3 +131,8 @@ exports.exportResults = (buildId, buildUrl) => {
     }
   });
 }
+
+exports.deleteResults = () => {
+  fs.unlink("log/build_results.txt", function (err){
+  });
+}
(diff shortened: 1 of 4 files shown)


nagpalkaran95 (approved):
lgtm

bin/helpers/utils.js
```diff
@@ -131,3 +131,8 @@ exports.exportResults = (buildId, buildUrl) => {
     }
   });
 }
+
+exports.deleteResults = () => {
+  fs.unlink("log/build_results.txt", function (err){
```
nagpalkaran95, line 136: add log line for error?

karanshah-browserstack: The unlink function throws an error even if there is no file present in the directory. And it is a validate case. Hence haven't logged any error.

nagpalkaran95 (approved):

nagpalkaran95 (approved):

PR #53, opened 2020-09-02
Environment variables support for better CI/CD integration

karanshah-browserstack:
Environment variables support for better CI/CD integration
Setting username , access_key, local and local identifiers if any env variable is set . 
Precedence of setting the keys : 
CLI > ENV_VAR > Config.json


SagarGaniga (approved):

prateekbstack:
LGTM

PR #54, opened 2020-09-02
Cypress config file support

suryart:
PR Implements following:

- Enable --ccf/--cypress-config-file options with `./cypress.json` as default
- The option provided enables user to set their preferred cypress.json config file or disable it altogether.
- Current implementation back port to `cypress_proj_dir` if both the values not set by users explicitly to avoid breaking the current flow.

Usage:

Absolute path:

    browserstack-cypress run --ccf /path/to/my.cypress.json

Relative path:

    browserstack-cypress run --ccf ./my.cypress.json

### DIFF
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -24,6 +24,9 @@ module.exports = function run(args) {
     // accept the build name from command line if provided
     utils.setBuildName(bsConfig, args);
 
+    // set cypress config filename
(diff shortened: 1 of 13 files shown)


SagarGaniga (dismissed):
lgtm

prateekbstack:
LGTM

SagarGaniga (approved):

PR #57, opened 2020-09-07
validateBstackJson & isCypressProjDirValid Fix

karanshah-browserstack:
Bug Fixes For : 
1. Adding stacktrace if browserstack.json is invalid
2. isCypressProjDirValid fails if you have pass integrationFoldDir is not an absolute Dir.


nagpalkaran95 (approved):

nagpalkaran95 (dismissed):

prateekbstack:
Tested for absolute and relative paths for both cypressProjectDir and integrationFolder. Works good with default cli dir as well as from any directory outside it with proper --config-file path for browserstack.json
Regression is pending, upon its completion, this will be QA approved

nagpalkaran95 (approved):

Issue #77, opened 2020-10-16
browserstack.json must contain auth object even if relying on environment variables


bdefore:
If relying on i.e. BROWSERSTACK_USERNAME environment variable, browserstack.json must contain `auth: {}` or else this error occurs: `Cannot set property 'username' of undefined`. Would be better if it worked as well without an empty object present.

Namrata-m13:
@bdefore A fix for this has been deployed. Let us know if it works for you as well.

bdefore:
i'm still seeing 1.5.1 as the latest in npm. is there another release you'd like me to try?

suryart:
Hey @bdefore,

This issue is addressed in the v1.5.2 release.

PR #79, opened 2020-10-26
Missing auth Hash causes exception in browserstack-cypress cli

karanshah-browserstack:
Fix for Missing auth Hash causes exception in browserstack-cypress cli.

### DIFF
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -17,6 +17,9 @@ module.exports = function run(args) {
   return utils.validateBstackJson(bsConfigPath).then(function (bsConfig) {
     utils.setUsageReportingFlag(bsConfig, args.disableUsageReporting);
 
+    // setting defaultAuthHash to {} if not present and set via env variables or via args.
+    utils.defaultAuthHash(bsConfig,args);
+
     // accept the username from command line or env variable if provided
     utils.setUsername(bsConfig, args);
 
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -17,8 +17,8 @@ module.exports = function run(args) {
   return utils.validateBstackJson(bsConfigPath).then(function (bsConfig) {
     utils.setUsageReportingFlag(bsConfig, args.disableUsageReporting);
 
-    // setting defaultAuthHash to {} if not present and set via env variables or via args.
-    utils.defaultAuthHash(bsConfig,args);
+    // setting setDefaultAuthHash to {} if not present and set via env variables or via args.
+    utils.setDefaultAuthHash(bsConfig,args);
 
     // accept the username from command line or env variable if provided
     utils.setUsername(bsConfig, args);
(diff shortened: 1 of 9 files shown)


bin/helpers/utils.js
```diff
@@ -116,6 +116,9 @@ exports.setParallels = (bsConfig, args) => {
 };
 
 exports.setUsername = (bsConfig, args) => {
+  if (this.isUndefined(bsConfig['auth']) && (!this.isUndefined(args.username) || !this.isUndefined(process.env.BROWSERSTACK_USERNAME))) {
```
suryart, line 119: We check and set it in validateBstackJson function instead since it is what loads the file and sents out the JSON back to all modules.

suryart:
Fixes #77

bin/helpers/utils.js
karanshah-browserstack: validateBstackJson function just checks if it is a validate json or not . After that we set this username and other variables set by using environment variables and post which the validations are performed.

suryart: Ok.

Why do we need an extra checks after && ? Why can't we just set the auth object if it is missing directly? Also, why not do it in setAccessKey?

Should we create another function in utils just for this that we use before setting username and access key? Thoughts?

karanshah-browserstack: If we don't have extra checks then , if the user doesn't pass any args or doesn't set via env vars the existing validation breaks and throws code error instead of validation error. 
We can create another function but would be an overhead as we already have this function .

suryart: >  We can create another function but would be an overhead as we already have this function .
Let's create another. Seem like a check/validation we should do separately and not involding username/accesskey domains.

bin/commands/runs.js
```diff
@@ -17,6 +17,9 @@ module.exports = function run(args) {
   return utils.validateBstackJson(bsConfigPath).then(function (bsConfig) {
     utils.setUsageReportingFlag(bsConfig, args.disableUsageReporting);
 
+    // setting defaultAuthHash to {} if not present and set via env variables or via args.
```
suryart, line 20: We should call this in other places, such as stop, info as well to make sure they're consistent as well.

karanshah-browserstack: Done.

suryart (dismissed):

bin/helpers/utils.js
```diff
@@ -115,6 +115,16 @@ exports.setParallels = (bsConfig, args) => {
   }
 };
 
+exports.defaultAuthHash = (bsConfig, args) => {
```
nopito, line 118: Hi there! What do you think about naming this as `setDefaultAuthValue`? So, this method uses the same format as others (action + noun), and also it uses a less agnostic type of data in the name?

karanshah-browserstack: Renamed it to setDefaultAuthHash

SagarGaniga (approved):

PR #82, opened 2020-10-30
Cli changes for sync cli 

karanshah-browserstack:
Implementation of CLI changes for sync and async cli .


bin/helpers/sync/syncSpecsLogs.js
```diff
@@ -105,6 +104,7 @@ let whileProcess = (whilstCallback) => {
         whileLoop = false;
         endTime = Date.now();
         showSpecsStatus(body);
+        logger.info("\n--------------------------------------------------------------------------------")
```
suryart, line 107: Why are we removing the above table config and adding static lines in logs here?

suryart (commented):
Rest LGTM

bin/helpers/sync/syncSpecsLogs.js
karanshah-browserstack: According to the spec , we don't need table config's border.

suryart (approved):

SagarGaniga (approved):

PR #87, opened 2020-11-05
Update logic for isCypressProjDirValid to handle windows path

nagpalkaran95:
* Fix for handling windows path in `integrationFolder` in cypress.json

### DIFF
--- a/test/unit/bin/helpers/utils.js
+++ b/test/unit/bin/helpers/utils.js
@@ -541,23 +541,28 @@ describe('utils', () => {
   });
 
   describe('isCypressProjDirValid', () => {
-    it('should return true when cypressDir and cypressProjDir is same', () => {
-      expect(utils.isCypressProjDirValid('/absolute/path', '/absolute/path')).to
-        .be.true;
+    it('should return true when cypressProjDir and integrationFoldDir is same', () => {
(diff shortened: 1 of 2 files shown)


test/unit/bin/helpers/utils.js
```diff
@@ -541,23 +541,28 @@ describe('utils', () => {
   });
 
   describe('isCypressProjDirValid', () => {
-    it('should return true when cypressDir and cypressProjDir is same', () => {
-      expect(utils.isCypressProjDirValid('/absolute/path', '/absolute/path')).to
-        .be.true;
+    it('should return true when cypressProjDir and integrationFoldDir is same', () => {
+      expect(utils.isCypressProjDirValid('/absolute/path', '/absolute/path')).to.be.true;
+
+      // should be as below for windows but path.resolve thinks windows path as a filename when run on linux/mac
```
karanshah-browserstack, line 547: If that's the case , do we need to comment this and add one more expect ?

karanshah-browserstack: This is for future use if we detect the platform is win32

karanshah-browserstack (approved):

suryart (approved):

prateekbstack:
LGTM

PR #91, opened 2020-11-10
[Sync CLI] Retry polling in case of network issue

SagarGaniga:
If there is some network issue while polling for data from our server, retry few times and handle the error properly with a descriptive error message.

### DIFF
--- a/bin/helpers/utils.js
+++ b/bin/helpers/utils.js
@@ -333,8 +333,8 @@ exports.handleSyncExit = (exitCode, dashboard_url) => {
 }
 
 exports.getNetworkErrorMessage = (dashboard_url) => {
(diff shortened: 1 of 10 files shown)


bin/helpers/utils.js
```diff
@@ -318,3 +321,20 @@ exports.setLocalIdentifier = (bsConfig) => {
 exports.capitalizeFirstLetter = (stringToCapitalize) => {
   return stringToCapitalize && (stringToCapitalize[0].toUpperCase() + stringToCapitalize.slice(1));
 };
+
+exports.handleSyncExit = (exitCode, dashboard_url) => {
```
karanshah-browserstack, line 325: Test cases for this function. ?

SagarGaniga: added

karanshah-browserstack (approved):
LGTM

PR #95, opened 2020-11-26
Multiple Versions of Cypress

SagarGaniga:
Adds a capability to provide cypress_version.

### DIFF
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -68,7 +68,8 @@ module.exports = function run(args) {
             }
 
             if (bsConfig.cypress_version && bsConfig.cypress_version !== data.cypress_version) {
-              logger.warn(Constants.userMessages.CYPRESS_VERSION_CHANGED);
+              message = utils.versionChangedMessage(bsConfig.cypress_version, data.cypress_version)
+              logger.warn(message);
             }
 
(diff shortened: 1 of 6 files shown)


bin/commands/runs.js
```diff
@@ -67,10 +67,15 @@ module.exports = function run(args) {
               logger.warn(Constants.userMessages.NO_PARALLELS);
             }
 
+            if (bsConfig.cypress_version && bsConfig.cypress_version !== data.cypress_version) {
+              logger.warn(Constants.userMessages.CYPRESS_VERSION_CHANGED);
```
suryart, line 71: let's check the message?

bin/helpers/utils.js
```diff
@@ -374,3 +374,9 @@ exports.getNetworkErrorMessage = (dashboard_url) => {
                   + Constants.userMessages.CHECK_DASHBOARD_AT  + dashboard_url
   return chalk.red(message)
 }
+
+exports.versionChangedMessage = (preferredVersion, actualVersion) => {
```
karanshah-browserstack, line 378: Can we add Spec for this?

bin/commands/runs.js
SagarGaniga: Updated

bin/helpers/utils.js
SagarGaniga: Done

karanshah-browserstack (approved):

prateekbstack:
LGTM

PR #97, opened 2020-12-07
return in case of error, improved error messaging

suryart:
Fix the error message and issues that occur due to loss of network connectivity.


karanshah-browserstack (approved):

prateekbstack:
LGTM

PR #103, opened 2021-01-04
Enable headless key in run_settings

suryart:
This PR enables the `headless` key to disable running browsers in headless mode on BrowserStack.

- When `headless` key is set to `false` or `"false"` it disabled the headless mode.
- When `headless` key is not provided or set to anything other than `false` or `"false"` it runs the tests in headless mode.
- Added `--headed` option to the `run` command which defaults to `true` and sets `headless` key as `false`.

### DIFF
--- a/bin/runner.js
+++ b/bin/runner.js
@@ -200,6 +200,11 @@ var argv = yargs
           default: false,
           describe: Constants.cliMessages.RUN.SYNC_DESCRIPTION,
           type: "boolean"
+        },
+        'headed': {
+          default: true,
+          describe: Constants.cliMessages.RUN.HEADED,
+          type: "boolean"
         }
       })
       .help('help')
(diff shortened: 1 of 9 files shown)


bin/runner.js
```diff
@@ -200,6 +200,11 @@ var argv = yargs
           default: false,
           describe: Constants.cliMessages.RUN.SYNC_DESCRIPTION,
           type: "boolean"
+        },
+        'headed': {
+          default: true,
```
karanshah-browserstack, line 205: The default should be set to false?

karanshah-browserstack:
Should we also update the default browserstack.json so users are aware of this option?

bin/runner.js
suryart: When `headed`, headless should be false, i.e. `headed` will be true.

karanshah-browserstack (dismissed):

SagarGaniga (approved):

prateekbstack:
LGTM

PR #107, opened 2021-01-11
Show error message when build is in running state

suryart:
In case when generate build command is fired when the build is running, this change will not show an error message instead of generating an incorrect build report.

### DIFF
--- a/bin/helpers/reporterHTML.js
+++ b/bin/helpers/reporterHTML.js
@@ -136,6 +136,8 @@ let reportGenerator = (bsConfig, buildId, args, cb) => {
         logger.info(message);
       }
     } else if (resp.statusCode === 422) {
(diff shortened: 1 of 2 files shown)


bin/helpers/reporterHTML.js
```diff
@@ -135,6 +135,15 @@ let reportGenerator = (bsConfig, buildId, args, cb) => {
         message = Constants.userMessages.API_DEPRECATED;
         logger.info(message);
       }
+    } else if (resp.statusCode === 422) {
```
SagarGaniga, line 138: Please add test cases for this

suryart: done

SagarGaniga (approved):

PR #112, opened 2021-02-11
fix: Update async whilst syntax

SagarGaniga:
`--sync` was failing because of the structural change in `async.whilst`

ref:
1.  https://stackoverflow.com/a/15393337/8266093
2. https://caolan.github.io/async/v3/docs.html#whilst


suryart (approved):

prateekbstack:
LGTM

PR #120, opened 2021-03-05
Wrap CSV input for --spec around with {}

SagarGaniga:
If a csv value is provided for --spec, it fails to detect the spec files correctly leading to incorrect parallel count calculation. This PR wraps the csv value around `{}` and if it is not csv then does nothing.

### DIFF
--- a/test/unit/bin/helpers/utils.js
+++ b/test/unit/bin/helpers/utils.js
@@ -1099,7 +1099,7 @@ describe('utils', () => {
       expect(utils.sanitizeSpecsPattern("pattern3")).to.eq("pattern3");
     });
 
-    it('should return null when --spec is undeffined', () => {
+    it('should return null when --spec is undefined', () => {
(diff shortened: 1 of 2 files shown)


test/unit/bin/helpers/utils.js
```diff
@@ -1089,6 +1089,21 @@ describe('utils', () => {
...
+
+    it('should wrap pattern around {} when input is csv', () => {
+      expect(utils.sanitizeSpecsPattern("pattern1,pattern2")).to.eq("{pattern1,pattern2}");
+    });
+
+    it('should not wrap pattern around {} when input is single glob pattern', () => {
+      expect(utils.sanitizeSpecsPattern("pattern3")).to.eq("pattern3");
+    });
+
+    it('should return null when --spec is undeffined', () => {
```
roshan04, line 1102: it should be undefined instead of undeffined.

roshan04 (approved):

karanshah-browserstack (approved):

prateekbstack:
LGTM

PR #121, opened 2021-03-08
Local mode and id validation

SagarGaniga:
Adds validation around local id and mode.
- local Identifier: Handle empty string
- local mode: handle boolean and other string values other than (on-demand / always on)

### DIFF
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -42,9 +42,6 @@ module.exports = function run(args) {
     //accept the local from env variable if provided
     utils.setLocal(bsConfig, args);
 
-    // set Local Mode (on-demand/ always-on)
-    utils.setLocalMode(bsConfig, args);
-
     //accept the local identifier from env variable if provided
     utils.setLocalIdentifier(bsConfig, args);
 
@@ -62,6 +59,9 @@ module.exports = function run(args) {
 
       // accept the number of parallels
       utils.setParallels(bsConfig, args, specFiles.length);
+
+      // set Local Mode (on-demand/ always-on)
+      utils.setLocalMode(bsConfig, args);
 
       // Archive the spec files
       return archiver.archive(bsConfig.run_settings, config.fileName, args.exclude).then(function (data) {
--- a/bin/helpers/utils.js
+++ b/bin/helpers/utils.js
@@ -578,6 +578,26 @@ exports.versionChangedMessage = (preferredVersion, actualVersion) => {
   return message
 }
 
+exports.invalidLocalMode = (bsConfig, args) => {
+  let localMode = undefined;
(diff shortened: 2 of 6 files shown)


bin/commands/runs.js
```diff
@@ -63,6 +60,9 @@ module.exports = function run(args) {
       // accept the number of parallels
       utils.setParallels(bsConfig, args, specFiles.length);
 
+      // set Local Mode (on-demand/ always-on)
+      utils.setLocalMode(bsConfig, args);
```
karanshah-browserstack, line 64: We depend on local mode for setting up local_identifier, I can still see local identifier getting set above.

bin/helpers/utils.js
```diff
@@ -586,6 +586,26 @@ exports.versionChangedMessage = (preferredVersion, actualVersion) => {
...
 }
 
+exports.invalidLocalMode = (bsConfig, args) => {
+  let localMode = undefined;
+  if (!this.isUndefined(args.localMode)) {
+    localMode = args.localMode;
+  } else if (!this.isUndefined(bsConfig.connection_settings)) {
+    localMode = bsConfig.connection_settings.local_mode
+  }
+  return (this.searchForOption('--local-mode') || !this.isUndefined(localMode)) && !["always-on","on-demand"].includes(localMode);
```
karanshah-browserstack, line 596: The condition is wrong. Please look at previous validation : 
Utils.searchForOption('--local-mode') && ( Utils.isUndefined(args.localMode) || (!Utils.isUndefined(args.localMode) && !["always-on","on-demand"].includes(args.localMode)))

bin/helpers/utils.js
```diff
@@ -586,6 +586,26 @@ exports.versionChangedMessage = (preferredVersion, actualVersion) => {
...
+}
+
+exports.invalidLocalIdentifier = (bsConfig, args) => {
+  let localId = undefined;
+  if (this.searchForOption('--local-identifier') && !this.isUndefined(args.localIdentifier)) {
```
karanshah-browserstack, line 601: if args.local identifier is undefined then we will throw and error.

bin/helpers/utils.js
```diff
@@ -586,6 +586,26 @@ exports.versionChangedMessage = (preferredVersion, actualVersion) => {
...
+    localId = !args.localIdentifier.toString().trim();
+  } else if (!this.isUndefined(bsConfig.connection_settings)) {
+    localId = !bsConfig.connection_settings.local_identifier.toString().trim();
```
karanshah-browserstack, line 604: what if local_identifier is undefined in bsConfig.connection_settings. This will give null pointer exception

karanshah-browserstack (changes requested):

PR #128, opened 2021-03-17
Adding unit test case and checking if local is passed as string

karanshah-browserstack:
Adding unit test case and checking if local is passed as string


SagarGaniga (approved):

prateekbstack:
LGTM

PR #136, opened 2021-04-15
text resizing based on the terminal column's width for sync logs

roshan04:
We will be widening or shorting the width of the column which shows the sync result logs for individual specs based on the current terminal's column width.

This PR adds the dynamically adjusting the specs listing column width based on the terminal/process' col size. Provides an option `--no-wrap` to disable the feature in case the user does not want the column to be resized.

### DIFF
--- a/bin/helpers/sync/syncSpecsLogs.js
+++ b/bin/helpers/sync/syncSpecsLogs.js
@@ -13,7 +13,7 @@ let specSummary = {
   "specs": [],
   "duration": null
 }
-let terminalWidth = (process.stdout.columns) * 0.9;
+let terminalWidth = (process.stdout.columns);
 
 let  getOptions = (auth, build_id) => {
   return {
@@ -34,8 +34,8 @@ let getTableConfig = () => {
   return {
     border: getBorderConfig(),
     columns: {
-      1: {alignment: 'center', width: 1},
-      2: {alignment: 'left', width: Math.floor(terminalWidth * 0.8)}
+      1: {alignment: 'center', width: Math.ceil(terminalWidth * 0.01)},
+      2: {alignment: 'left', width: Math.floor(terminalWidth * 0.70)}
     },
     columnDefault: {
       width: Math.floor(terminalWidth * 0.2),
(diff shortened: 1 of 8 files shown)


bin/helpers/sync/syncSpecsLogs.js
```diff
@@ -13,7 +13,8 @@ let specSummary = {
   "specs": [],
   "duration": null
 }
-
+let noWrap = utils.searchForOption('--no-wrap')
```
SagarGaniga, line 16: can we add validation around `--no-wrap`?

bin/runner.js
```diff
@@ -221,6 +221,9 @@ var argv = yargs
         'local-config-file': {
           describe: Constants.cliMessages.RUN.LOCAL_CONFIG_FILE,
           type: "string"
+        },
+        'no-wrap': {
```
SagarGaniga, line 225: add a default value, type, and alias

bin/helpers/sync/syncSpecsLogs.js
```diff
@@ -81,7 +82,7 @@ let printSpecsStatus = (bsConfig, buildDetails) => {
         whileProcess(callback)
       },
       function(err, result) { // when loop ends
-        logger.info("\n--------------------------------------------------------------------------------")
+        noWrap ? logger.info("\n--------------------------------------------------------------------------------") : logger.info("\n" + "-".repeat(terminalWidth))
```
SagarGaniga, line 85: can we use `noWrap?` to determine separator length in both cases? Something like
```
logger.info("-".repeat(noWrap ? X : terminalWidth))
```

suryart: The `X` is default and we would want the existing behaviour to continue in case of a `no-wrap`. We can make it something like:
```node
logger.info("-".repeat(noWrap ? 80 : terminalWidth))
```
Where 80 is the current number of the `-` which are being printed.

bin/runner.js
suryart: Done

SagarGaniga (approved):
LGTM

SagarGaniga (approved):

prateekbstack:
LGTM

PR #143, opened 2021-06-07
Added metrics for cypress local setup time

sayak119:
Added phase 2 cypress metrics

### DIFF
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -77,12 +77,13 @@ module.exports = function run(args) {
           // Create build
 
           //setup Local Testing
+          let options = {};
           let startTime = new Date().getTime();
           let bs_local = await utils.setupLocalTesting(bsConfig, args);
-          let endTime = new Date().getTime();
-          let setupTimeForLocalTesting = (endTime - startTime)/1000;
+          let endTime = new Date().getTime();          
+          options.setupTimeForLocalTesting = (endTime - startTime)/1000;
 
-          return build.createBuild(bsConfig, zip, setupTimeForLocalTesting).then(function (data) {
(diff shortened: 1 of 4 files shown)


bin/commands/runs.js
```diff
@@ -77,9 +77,12 @@ module.exports = function run(args) {
           // Create build
 
           //setup Local Testing
+          let startTime = new Date().getTime();
           let bs_local = await utils.setupLocalTesting(bsConfig, args);
+          let endTime = new Date().getTime();
+          let setupTimeForLocalTesting = (endTime - startTime)/1000;
 
-          return build.createBuild(bsConfig, zip).then(function (data) {
+          return build.createBuild(bsConfig, zip, setupTimeForLocalTesting).then(function (data) {
```
suryart, line 85: Let's push the time in an `options` Object instead of sending it as a param.

suryart (changes requested):

bin/commands/runs.js
```diff
@@ -77,9 +77,13 @@ module.exports = function run(args) {
           // Create build
 
           //setup Local Testing
+          let options = {};
+          let startTime = new Date().getTime();
           let bs_local = await utils.setupLocalTesting(bsConfig, args);
```
suryart, line 82: This will be returning [a promise](https://github.com/browserstack/browserstack-cypress-cli/blob/master/bin/helpers/utils.js#L445) so checking the start and end time and taking the diff will not really help. We should check for [actual start time](https://github.com/browserstack/browserstack-cypress-cli/blob/master/bin/helpers/utils.js#L450-L455) for loggin.

suryart (changes requested):

francisf:
outdated. will re-visit.

PR #145, opened 2021-06-21
add session time components instrumentation

shashikdm:
Added session time components file. sending this data to cli stats, honouring the user's permission to send usage stats

### DIFF
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -68,7 +68,7 @@ module.exports = function run(args) {
     // Validate browserstack.json values and parallels specified via arguments
     markBlockStart('validateConfig');
     return capabilityHelper.validate(bsConfig, args).then(function (cypressJson) {
-
+      markBlockEnd('validateConfig');
       markBlockStart('preArchiveSteps');
       //get the number of spec files
       let specFiles = utils.getNumberOfSpecFiles(bsConfig, args, cypressJson);
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -129,8 +129,19 @@ module.exports = function run(args) {
             logger.info(message);
             logger.info(dashboardLink);
             if(!args.sync) logger.info(Constants.userMessages.EXIT_SYNC_CLI_MESSAGE.replace("<build-id>",data.build_id));
-            utils.sendUsageReport(bsConfig, args, `${message}\n${dashboardLink}`, Constants.messageTypes.SUCCESS, null);
-            utils.sendUsageReport(bsConfig, args, `buildId: ${data.build_id}`, Constants.messageTypes.TIMECOMPONENTS, null, getTimeComponents());
+            let dataToSend = {
+              time_components: getTimeComponents(),
+              build_id: data.build_id,
+            };
+            if (bsConfig && bsConfig.connection_settings) {
+              if (bsConfig.connection_settings.local_mode) {
+                dataToSend.local_mode = bsConfig.connection_settings.local_mode;
+              }
+              if (bsConfig.connection_settings.usedAutoLocal) {
+                dataToSend.used_auto_local = bsConfig.connection_settings.usedAutoLocal;
+              }
+            }
+            utils.sendUsageReport(bsConfig, args, `${message}\n${dashboardLink}`, Constants.messageTypes.SUCCESS, null, dataToSend);
             return;
           }).catch(async function (err) {
             // Build creation failed
--- a/bin/commands/runs.js
+++ b/bin/commands/runs.js
@@ -130,7 +130,7 @@ module.exports = function run(args) {
             logger.info(dashboardLink);
             if(!args.sync) logger.info(Constants.userMessages.EXIT_SYNC_CLI_MESSAGE.replace("<build-id>",data.build_id));
             utils.sendUsageReport(bsConfig, args, `${message}\n${dashboardLink}`, Constants.messageTypes.SUCCESS, null);
-            utils.sendUsageReport(bsConfig, args, 'Time Components', Constants.messageTypes.TIMECOMPONENTS, null, getTimeComponents());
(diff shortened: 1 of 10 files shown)


suryart (dismissed):

ojasrahatebs:
REgression Approved!

bin/helpers/timeComponents.js
```diff
@@ -0,0 +1,65 @@
...
+
+let sessionTimes = {
+  referenceTimes: {
+    absoluteStartTime: Date.now()
+  }, // Absolute times which needs to be used later to calculate logTimes
+  logTimes: {}, // Time Difference in ms which we need to push to EDS
+};
+
+const initTimeComponents = () => {
+  sessionTimes = {referenceTimes: {absoluteStartTime: Date.now()}, logTimes: {}};
```
karanshah-browserstack, line 13: Why are we duplicating it? It is already initialized on line number 5?

bin/commands/runs.js
```diff
@@ -9,14 +9,21 @@ const archiver = require("../helpers/archiver"),
...
   fileHelpers = require("../helpers/fileHelpers"),
   syncRunner = require("../helpers/syncRunner"),
-  reportGenerator = require('../helpers/reporterHTML').reportGenerator;
+  reportGenerator = require('../helpers/reporterHTML').reportGenerator,
+  {initTimeComponents, markBlockStart, markBlockEnd, getTimeComponents} = require('../helpers/timeComponents');
 
 module.exports = function run(args) {
   let bsConfigPath = utils.getConfigPath(args.cf);
   //Delete build_results.txt from log folder if already present.
+  initTimeComponents();
```
karanshah-browserstack, line 18: Not needed according to current implementation

bin/commands/runs.js
```diff
@@ -9,14 +9,21 @@ const archiver = require("../helpers/archiver"),
...
+  markBlockStart('deleteOldResults');
   utils.deleteResults();
+  markBlockEnd('deleteOldResults');
 
+  markBlockStart('validateBstackJson');
   return utils.validateBstackJson(bsConfigPath).then(function (bsConfig) {
+    markBlockEnd('validateBstackJson');
```
karanshah-browserstack, line 25: What if the validateBstackJson throws an error?

bin/commands/runs.js
```diff
@@ -56,10 +63,13 @@ module.exports = function run(args) {
 
     // set the no-wrap
     utils.setNoWrap(bsConfig, args);
+    markBlockEnd('setConfig');
 
     // Validate browserstack.json values and parallels specified via arguments
+    markBlockStart('validateConfig');
```
karanshah-browserstack, line 69: Where are we ending it?

bin/commands/runs.js
```diff
@@ -69,17 +79,29 @@ module.exports = function run(args) {
...
 
           //setup Local Testing
+          markBlockStart('localSetup');
           let bs_local = await utils.setupLocalTesting(bsConfig, args);
-
+          markBlockEnd('localSetup');
+          markBlockStart('createBuild');
           return build.createBuild(bsConfig, zip).then(function (data) {
+            markBlockEnd('createBuild');
+            markBlockEnd('total');
```
karanshah-browserstack, line 104: Where have we started the block for total?

karanshah-browserstack (changes requested):

bin/commands/runs.js
shashikdm: The internal logic handles this. it'll take from creation of the object

bin/helpers/timeComponents.js
shashikdm: Which is fine for current case. But going forward we may need explicit initialization hence keeping this.

bin/commands/runs.js
shashikdm: this is to explicitly state that we're recording from here

bin/commands/runs.js
shashikdm: we do not want time measurements for unsuccessful builds as of now.

bin/commands/runs.js
shashikdm: thanks for pointing out. adding markBlockEnd for this.

shashikdm: this is done in latest commit

bin/helpers/timeComponents.js
karanshah-browserstack: Can we please do it now?

bin/commands/runs.js
karanshah-browserstack: Then can we remove the extra initialization?

bin/commands/runs.js
karanshah-browserstack: Is this in sync with the product?

bin/commands/runs.js
karanshah-browserstack: Any issues in explicit initialization?

shashikdm: no issue as such, it'll be redundant.

bin/commands/runs.js
shashikdm: yes

bin/helpers/timeComponents.js
shashikdm: explicitly initializing in like 18 run.js.

bin/commands/runs.js
shashikdm: The global initialization is put in place so that the function calls never break the flow and cause runtime error since this is purely instrumentation code. `initTimeComponents` is added here to start measuring time from this point.

bin/commands/runs.js
karanshah-browserstack: Then please remove it.

karanshah-browserstack (approved):

PR #148, opened 2021-07-20
Add support for checking if zip already uploaded to browserstack.

sparsh00:
Added support to check if zip already uploaded by checking md5 of the zip files.


bin/helpers/constants.js
```diff
@@ -145,6 +147,21 @@ const filesToIgnoreWhileUploading = [
   '.yarn/**',
 ];
 
+const readDirOptions = {
+  cwd: '.',
```
karanshah-browserstack, line 151: Why are we not using cypressFolderPath as CWD? We have the user's who have their cypress.json and CWD different.

karanshah-browserstack (changes requested):

bin/helpers/constants.js
sparsh00: Will be using that only https://github.com/browserstack/browserstack-cypress-cli/pull/148/files#diff-e4adc3c50732024d070d2df81b167010ada80e39e52edfc8f3f48b0e9b44a501R18 https://github.com/browserstack/browserstack-cypress-cli/pull/148/files#diff-a91b9660eaaccd540cc602da6acb2b2f1bcad2dc63d25de89bd569463295d57eR20 this is just for fallback.

bin/helpers/checkUploaded.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+const request = require('request');
+
+const crypto = require('crypto'),
```
shashikdm, line 4: this is a new dependency. we may need to add this in `package.json`

shashikdm: ignore. its part of node js standard library

shashikdm (approved):

karanshah-browserstack (approved):

shashikdm (approved):

karanshah-browserstack (approved):

PR #149, opened 2021-07-27
Changes in CLI messages to support browserstack cypress latest flag

pranavj1001:
This PR adds the following changes
1. adds a new info message which is displayed when a user passes the new BrowserStack Cypress latest syntax.
2. hides the warning message which was displayed earlier when the user's passed version did not match with the BrowserStack supported Cypress version if the user passes the new BrowserStack Cypress latest syntax.


pranavj1001:
Created a new one https://github.com/browserstack/browserstack-cypress-cli/pull/156 directly on this repo, closing this one.

PR #159, opened 2021-09-02
Improved support of config options in cli

karanshah-browserstack:
Support of different config options provided by cypress.

1. --browser
2. --config
3. --parallel
4. --ci-build-id

### DIFF
--- a/bin/helpers/utils.js
+++ b/bin/helpers/utils.js
@@ -739,3 +739,32 @@ exports.deleteBaseUrlFromError = (err) => {
   return err.replace(/To test ([\s\S]*)on BrowserStack/g, 'To test on BrowserStack');
 }
 
+exports.setBrowsers = async (bsConfig, args) => {
+  return new Promise((resolve, reject) => {
+    if(!this.isUndefined(args.browser)){
+      try{
+        bsConfig["browsers"] = []
+        let browsersList = args.browser.split(',')
+        browsersList.forEach((browser)=>{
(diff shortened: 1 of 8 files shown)


bin/helpers/utils.js
```diff
@@ -745,3 +745,32 @@ exports.deleteBaseUrlFromError = (err) => {
   return err.replace(/To test ([\s\S]*)on BrowserStack/g, 'To test on BrowserStack');
 }
 
+exports.setBrowsers = async (bsConfig, args) => {
+  return new Promise((resolve, reject) => {
+    if(!this.isUndefined(args.browser)){
+      try{
+        bsConfig["browsers"] = []
+        let browsersList = args.browser.split(',')
+        browsersList.forEach((browser)=>{
```
nagpalkaran95, line 754: convert to simple for loop as forEach doesn't work with async await

nagpalkaran95 (changes requested):

nagpalkaran95 (approved):

bin/helpers/utils.js
karanshah-browserstack: If there is an async-await inside forEach it doesn't work. By default, it does work fine.

nagpalkaran95 (dismissed):

nagpalkaran95 (approved):

rahulpsq (commented):
LGTM

PR #161, opened 2021-09-14
Version bump

suryart:
Version bump


kunal361 (approved):

.npmignore
```diff
@@ -0,0 +1,2 @@
+test
```
karanshah-browserstack, line 1: Why test folder?

nagpalkaran95 (approved):

.npmignore
suryart: We don't need users to download test dir when running `npm install`

PR #182, opened 2021-12-08
Instrument zip sizes

nagpalkaran95:


test/unit/bin/commands/runs.js
```diff
@@ -821,13 +837,15 @@ describe("runs", () => {
...
       checkErrorStub.returns('test');
+      fetchZipSizeStub.returns(123);
       createBuildStub.returns(Promise.resolve({ message: 'Success', build_id: 'random_build_id', dashboard_url: dashboardUrl }));
 
       return runs(args)
         .then(function (_bsConfig) {
           chai.assert.fail("Promise error");
         })
         .catch((error) => {
+          console.log(error);
```
sagar-kh3, line 848: Pls remove unless it's deliberate

bin/helpers/utils.js
```diff
@@ -1012,3 +1012,13 @@ async function processExitHandler(exitData){
...
   process.exit(0);
 }
+
+exports.fetchZipSize = (fileName) => {
+  try {
+    let stats = fs.statSync(fileName)
+    return stats.size; // in bytes
+  }
+  catch(err) {
+    return "file not present"
```
sagar-kh3, line 1022: I'd suggest to keep this 0 to have consistency of data type (for ex, someone running a P90 or other mathematical function on size of zip on BQ would expect size to be always a number and not string).

sagar-kh3 (changes requested):

PR #191, opened 2021-12-16
instrument zip upload time split

nagpalkaran95:


bin/commands/runs.js
```diff
@@ -201,6 +202,14 @@ module.exports = function run(args, rawArgs) {
                   test_zip_size: test_zip_size,
                   npm_zip_size: npm_zip_size,
                 };
+
+                if (zip.tests_upload_time || zip.npm_package_upload_time) {
+                  dataToSend.time_components.zip.zipUploadSplit = {
```
roshan04, line 207: any specific reason to add upload time keys in such nested format cant we keep them just under zip, if there is some reason then lets keep zipUploadSplit in the same snake case naming style as other keys are.

nagpalkaran95: Post the change the obj would like this:
```
zip: {
      archive: 173,
      zipUpload: 1297,
      total: 1470,
      zipUploadSplit: { tests_upload_time: 1297, npm_package_upload_time: 1000 }
    }
```
where `zipUpload` gives the total time of uploading and zipUploadSplit which gives the split. Also to add, both packages are uploaded in parallel so you can't just add the split numbers to get the total time but it would be the max of both.

roshan04: cool, then changes looks good to me.

roshan04 (approved):
LGTM

pranavj1001 (approved):

PR #194, opened 2021-12-20
replacing axios to native request to handle proxy issue

roshan04:

PR #183, opened 2021-12-09
Add progress bar with ETA, speed and duration for zip uploads

nagpalkaran95:
image

PR #196, opened 2021-12-27
Remove dependency on axios and add progress bar

nagpalkaran95:


pranavj1001 (approved):
Approving this as it is a merge of https://github.com/browserstack/browserstack-cypress-cli/pull/183 and https://github.com/browserstack/browserstack-cypress-cli/pull/194. And both of these PRs are already approved.

pranavj1001 (approved):

PR #206, opened 2022-01-07
handled case for agnostic sessions with browser flag

roshan04:


nagpalkaran95:
https://github.com/browserstack/browserstack-cypress-cli/blob/master/bin/helpers/capabilityHelper.js#L26
handling required here?

nagpalkaran95 (approved):

PR #221, opened 2022-01-17
Fix "Error: spawn npm ENOENT" on windows

viszkoktamas:
I was running this lib on windows with a few npm_dependencies and i got this spawn npm error.
asd

Fixed with the solution posted on SO: https://stackoverflow.com/a/43285131/11882999

### DIFF
--- a/bin/helpers/packageInstaller.js
+++ b/bin/helpers/packageInstaller.js
@@ -63,7 +63,7 @@ const packageInstall = (packageDir) => {
       logger.error(`Some error occurred while installing packages: ${error}`);
       reject(`Packages were not installed successfully.`);
     };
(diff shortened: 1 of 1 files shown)


shawnlobo96:
@viszkoktamas what version of the CLI are you using?

This issue was fixed in `1.11.2`. 

If you're using something older than `1.11.2`, can you give it a try with the latest CLI version?

viszkoktamas:
@shawnlobo96 I've just upgraded to 1.11.2, now the cli doesn't fails on the error, but the problem is still exists, check this screenshot:
image

Test run fails on missing dependency of course (the error message is a bit confusing, but same setup works using Ubuntu, and the error message in the first screenshot implies that it is caused by the npm spawn error):
image

nagpalkaran95 (approved):

PR #247, opened 2022-02-16
Add verbose logging

07souravkunda:


bin/helpers/buildArtifacts.js
```diff
@@ -229,6 +230,7 @@ exports.downloadBuildArtifacts = async (bsConfig, buildId, args, rawArgs) => {
             await sendUpdatesToBstack(bsConfig, buildId, args, options, rawArgs)
             utils.sendUsageReport(bsConfig, args, message, messageType, null, null, rawArgs);
           }
+          logger.debug(JSON.stringify(resp));
```
pranavj1001, line 233: Beware of circular references in the `resp` object. `JSON.stringify` throws an error in case of a circular reference. An alternative for this is to use `util.inspect`. Before using this do check the performance difference between the two.

pranavj1001 (dismissed):

PR #284, opened 2022-03-22
Support for record flag

roshan04:


bin/helpers/capabilityHelper.js
```diff
@@ -241,6 +241,14 @@ const validate = (bsConfig, args) => {
       logger.warn(Constants.validationMessages.SPEC_TIMEOUT_NOT_PASSED_ERROR);
     }
 
+    if(!Utils.isUndefined(bsConfig.run_settings["record"])) {
```
pranavj1001, line 244: if `bsConfig.run_settings["record"]` is false then I believe, it will go inside this code block and give warnings to users if the project Id or record key is missing which it ideally shouldn't

bin/helpers/capabilityHelper.js
```diff
@@ -241,6 +241,14 @@ const validate = (bsConfig, args) => {
...
+      if(Utils.isUndefined(bsConfig.run_settings.projectId)) {
+        logger.warn(Constants.validationMessages.PROJECT_ID_MISSING);
+      } else if (Utils.isUndefined(bsConfig.run_settings["record-key"])) {
```
pranavj1001, line 247: Let's make this a separate `if` condition and not `else if`. The reason for this is to let users know on the first try only they are missing both projectId and record-key if they are missing.

bin/helpers/utils.js
```diff
@@ -327,6 +327,39 @@ exports.setSpecTimeout = (bsConfig, args) => {
...
+  } else if (!this.isUndefined(process.env.CYPRESS_RECORD_KEY)) {
+    return process.env.CYPRESS_RECORD_KEY;
+  }
+  return bsConfig.run_settings["record-key"];
+}
+
+exports.setProjectId = (bsConfig, args) => {
+  if(!this.isUndefined(args["projectId"])) {
+    return args["projectId"];
+  } else if(!this.isUndefined(bsConfig.run_settings["projectId"])) {
```
pranavj1001, line 349: Just like env var `CYPRESS_RECORD_KEY`, we have an env var for project id i.e. `CYPRESS_PROJECT_ID`. More info is available [here](https://docs.cypress.io/guides/dashboard/projects#Project-ID).

bin/helpers/capabilityHelper.js
```diff
@@ -241,6 +241,15 @@ const validate = (bsConfig, args) => {
       logger.warn(Constants.validationMessages.SPEC_TIMEOUT_NOT_PASSED_ERROR);
     }
 
+    if(!Utils.isUndefined(bsConfig.run_settings["record"]) && String(bsConfig.run_settings["record"]) == 'true') {
+      if(Utils.isUndefined(bsConfig.run_settings.projectId)) {
```
pranavj1001, line 245: let's add check for `projectId` or `record-key` being empty strings as well and show the warning.

pranavj1001 (dismissed):

nagpalkaran95 (dismissed):
recent changes LGTM

nagpalkaran95 (approved):

pranavj1001 (approved):

PR #295, opened 2022-04-04
Add code to output custom error messages on CLI from backend

pranavj1001:


bin/helpers/sync/specsSummary.js
```diff
@@ -32,6 +33,24 @@ let printSpecsRunSummary = (data, machines) => {
...
+            winstonlogger.info(error.message);
+            break;
+          case 'warn':
+            winstonlogger.warn(error.message);
+            break;
+          case 'error':
+            winstonlogger.error(error.message);
+            break;
+          default:
+            winstonlogger.warn(error.message);
```
nagpalkaran95, line 49: can remove
```
case 'warn':
            winstonlogger.warn(error.message);
            break;
```

It will be taken care of in default block

nagpalkaran95 (changes requested):

bin/helpers/sync/specsSummary.js
pranavj1001: Done

nagpalkaran95 (approved):

PR #312, opened 2022-04-21
Improving instrumentation

roshan04:
Track user_id in cli stats for failures before build creation.
https://browserstack.atlassian.net/browse/CYP-1119


bin/commands/runs.js
```diff
@@ -141,20 +148,16 @@ module.exports = function run(args, rawArgs) {
...
-                utils.setProcessHooks(data.build_id, bsConfig, bs_local, args);
+                utils.setProcessHooks(data.build_id, bsConfig, bs_local, args, buildReportData);
                 let message = `${data.message}! ${Constants.userMessages.BUILD_CREATED} with build id: ${data.build_id}`;
                 let dashboardLink = `${Constants.userMessages.VISIT_DASHBOARD} ${data.dashboard_url}`;
-                let buildReportData = {
-                  'build_id': data.build_id,
-                  'user_id': data.user_id,
-                  'parallels': userSpecifiedParallels
-                };
+                buildReportData = { 'build_id': data.build_id, 'user_id': data.user_id, ...buildReportData }
```
roshan04, line 160: should merge parallels instead of user_id

bin/commands/stop.js
```diff
@@ -18,14 +19,20 @@ module.exports = function stop(args, rawArgs) {
     // accept the access key from command line if provided
     utils.setAccessKey(bsConfig, args);
 
+    let initDetails = await getInitialDetails(bsConfig, args, rawArgs);
```
roshan04, line 22: lets store the result after first API call if data is not available then hit the api. Lets make this code section common under a function.

pranavj1001 (approved):

PR #341, opened 2022-06-02
Json html report changes

roshan04:
Download report json and html.

### DIFF
--- a/bin/helpers/reporterHTML.js
+++ b/bin/helpers/reporterHTML.js
@@ -103,12 +103,7 @@ async function generateCypressBuildReport(report_data) {
     logger.debug("Creating results directory.");
     fs.mkdirSync(resultsDir);
   }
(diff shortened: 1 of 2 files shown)


bin/helpers/reporterHTML.js
```diff
@@ -5,88 +5,7 @@ const fs = require('fs'),
...
-    specStats.label = `${failedSpecs}/${totalSpecs} FAILED`;
-    specStats.cssClass = 'spec-stats-failed';
-  } else {
-    specStats.label = `${passedSpecs}/${totalSpecs} PASSED`;
-    specStats.cssClass = 'spec-stats-passed';
-  }
-
-  return specStats;
-}
+const unzipper = require('unzipper');
```
pranavj1001, line 8: let's add this import statement below line no. 3

pranavj1001:
As per the new debug flow, let's add more logs to the code at every important step. We should make it a bit more verbose.

pranavj1001 (dismissed):

PR #420, opened 2022-10-14
Set default spec folder depending on the test suite type

pranavj1001:


roshan04 (approved):

droca:
Any chance to get this merged anytime soon?

melibe23:
We would really need this fix merged. We are using Cypress v12 and we cannot parallelize even if we are paying BS to use up to 50 paral.

PR #441, opened 2022-11-29
Ignore tests.zip from uploads when home_directory is set

### LINKED ISSUE #440: Recursion creating tests.zip when home_directory is defined
When `home_directory` is set to a parent directory, `tests.zip` ends up including itself while being generated, which ends up in a recursion and an infinitely growing file.

ismaelgt:
Adds `**/tests.zip` to the list of patterns ignored for uploading. 

This avoids recursion creating `tests.zip` when `home_directory` is defined.

Closes #440


suryart (approved):
LGTM

PR #533, opened 2023-03-06
Support cypress network logs

roshan04:


bin/helpers/utils.js
```diff
@@ -1045,6 +1045,14 @@ exports.getNetworkErrorMessage = (dashboard_url) => {
   return chalk.red(message)
 }
 
+exports.setNetworkLogs = (bsConfig) => {
+  if(bsConfig.run_settings.networkLogs == 'true' || bsConfig.run_settings.networkLogs == true) {
```
pranavj1001, line 1049: Are we adding support for `networkLogs` as a CLI arg?

roshan04: No as per product spec we will only be supporting networkLogs in browserstack.json

roshan04: https://browserstack.atlassian.net/wiki/spaces/PROD/pages/3658908037/Cypress+Network+Logs#Requirements

pranavj1001 (approved):

PR #585, opened 2023-05-19
:sparkle: added timezone support

Archish27:
Pass `timezone` capability in `run_settings` in `browserstack.json` & the values will be as per the timezone document - [Configure timezones for Selenium tests | BrowserStack Docs](https://www.browserstack.com/docs/automate/selenium/configure-timezones)

### DIFF
--- a/bin/helpers/utils.js
+++ b/bin/helpers/utils.js
@@ -395,12 +395,14 @@ exports.isValidTimezone = (timezone) => this.isNotUndefined(timezone) && this.is
 exports.setTimezone = (bsConfig, args) => {
   let timezone = args.timezone || bsConfig.run_settings.timezone;
   let newTimezone;
-  if(this.isValidTimezone(timezone)){
-    newTimezone = timezone; 
-  } else {
-    logger.error(`Invalid timezone = ${timezone}`);
-    syncCliLogger.info(chalk.red(Constants.userMessages.INVALID_TIMEZONE));
-    process.exit(1);
+  if(this.isNotUndefined(timezone)) {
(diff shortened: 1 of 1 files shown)


bin/helpers/utils.js
```diff
@@ -389,6 +390,42 @@ exports.setSpecTimeout = (bsConfig, args) => {
...
+exports.isValidTimezone = (timezone) => {
+  if(!this.isUndefined(timezone) && !this.isUndefined(TIMEZONE[timezone])) {
+    return true;
+  }
+  return false;
+}
+
+exports.setTimezone = (bsConfig, args) => {
+  let timezone = undefined;
+  if(this.isTimezoneArgPassed()) {
```
roshan04, line 406: we can take a variable and save the timezone value passed with OR condition from args and browserstack.json and check check for validations. This will help to reduce redudant code

bin/helpers/constants.js
```diff
@@ -85,6 +85,8 @@ const userMessages = {
     "Your build will run using Cypress <actualVersion> instead of Cypress <preferredVersion>.<frameworkUpgradeMessage> Read more about supported versions here: http://browserstack.com/docs/automate/cypress/supported-versions",
   LOCAL_START_FAILED: "Local Testing setup failed.",
   LOCAL_STOP_FAILED: "Local Binary stop failed.",
+  INVALID_TIMEZONE:
```
roshan04, line 88: Verified by product ??

roshan04 (dismissed):

roshan04 (approved):

PR #596, opened 2023-06-12
❇️ migration from request library to axios

Archish27:


xcafebabe:
Hi @Archish27 ! I'm glad you're making this change!, Is there a way we can build the cli with your branch version  to test it in our pipelines, in that way we can also help as beta testers.

Archish27:
@xcafebabe Yes, it's working branch. Just it doesn't have newer features which were released in cypress cli. Still you can use this branch to trigger tests on BrowserStack.

xcafebabe:
> @xcafebabe Yes, it's working branch. Just it doesn't have newer features which were released in cypress cli. Still you can use this branch to trigger tests on BrowserStack.

ok cool! i'm going to use this branch in our pipelines 👍🏼

xcafebabe:
so far so good 👍🏼

pallost:
hi, any news on this? The "request" library is causing audit issues, so a migration to axios is more than welcome

droca:
Is this development abandoned?

PR #671, opened 2023-08-17
fix: Git mapping for windows

sauravdas1997:


amaanbs (approved):

bin/testObservability/helper/helper.js
```diff
@@ -163,7 +163,7 @@ exports.getTestEnv = () => {
 }
 
 exports.getFileSeparatorData = () => {
-  const fileSeparatorRegex = /^win/.test(process.platform) ? "\\\\" : "/";
+  const fileSeparatorRegex = /^win/.test(process.platform) ? "\\" : "/";
```
amaanbs, line 166: Now that `fileSeparatorRegex` & `fileSeparator` are the same values, let's refactor to send a single string and use that for all file separator operations.

amaanbs (changes requested):

amaanbs (approved):

asambstack (approved):
Discussed with @sauravdas1997 
`\\\\` was added by another dev based on some SO threads and has now been removed and tested.
LGTM

PR #779, opened 2024-02-15
Add support for ccf flag for O11y

amaanbs:


bin/commands/runs.js
```diff
@@ -61,6 +61,9 @@ module.exports = function run(args, rawArgs) {
     markBlockStart('setConfig');
     logger.debug('Started setting the configs');
 
+    // set cypress config filename
+    utils.setCypressConfigFilename(bsConfig, args);
```
karanshah-browserstack, line 65: Let's move our code in end in this scenario?

karanshah-browserstack (commented):
Can see this can impact other stuff too

07souravkunda (approved):

bin/commands/runs.js
amaanbs: That would require moving more code around. This change shouldn't affect handling of other products.

rev-doshi (commented):
This same changes is also needed for A11y AFAIR, can you test that once ?

amaanbs:
> This same changes is also needed for A11y AFAIR, can you test that once ?
Tested, working fine.

RutvikChandla (approved):

pranavj1001 (approved):

PR #841, opened 2024-07-02
Bump the version of update-notifier package

asambstack:


sparsh00 (approved):
Lets just verify that moving to async doesn't cause it to interfere with other flows

PR #855, opened 2024-08-22
[SDK-1040] reduce batch interval

samarsault:


bin/testObservability/helper/constants.js
```diff
@@ -2,7 +2,7 @@ const path = require('path');
 
 exports.consoleHolder = Object.assign({},console);
 exports.BATCH_SIZE = 1000;
-exports.BATCH_INTERVAL = 2000;
+exports.BATCH_INTERVAL = 1000;
```
sauravdas1997, line 5: we do not want to reduce this.
- Increase the teardown time of request batcher
- Add more debug logging to track the flow of the events.

PR #867, opened 2024-09-08
Add changes for splitting mac and win config

asambstack:


sparsh00:
One optimization we can do is check user OS and do local NPM installation only on base + OS-specific packages. This will ensure that the local npm install works, plus npm rebuild will also work for that specific os

sparsh00 (approved):

PR #962, opened 2025-02-06
Browserstack Accessibility - Fix duplicate queries error

amaanbs:


bin/accessibility-automation/cypress/index.js
```diff
@@ -424,7 +424,11 @@ Cypress.Commands.add('getAccessibilityResults', () => {
 	}
 });
 
-Cypress.Commands.addQuery('performScanSubjectQuery', function (chaining, setTimeout) {
-    this.set('timeout', setTimeout);
-    return () => cy.getSubjectFromChain(chaining);
-});
+if (!Cypress.Commands.hasOwnProperty('_myQueryAdded')) {
```
yashdsaraf, line 427: Can we make this a bit more context specific, maybe something like
```suggestion
if (!Cypress.Commands.hasOwnProperty('_browserstackSDKQueryAdded')) {
```

amaanbs: Done.

sauravdas1997 (approved):

PR #986, opened 2025-08-01
Fix cypress v13 approach 2

07souravkunda:


amaanbs (dismissed):

osho-20 (dismissed):

karanshah-browserstack:
Ensure testing on all cypress versions with testObservability as true and false without their imports too. 
Also, ensure testing of A11y for these versions. 
If everything looks good, please share a GTG from QA. We should be good.

karanshah-browserstack:
> @karanshah-browserstack Please confirm which versions are expected to be covered in QA automation.

Please check all the versions which are supported atleast the least supported one, v10, v12, v13 & v14

PR #988, opened 2025-08-05
Fix: Double Ally testcase save calls 

Bhargavi-BS:


bin/helpers/helper.js
```diff
@@ -390,7 +390,7 @@ exports.getSupportFiles = (bsConfig, isA11y) => {
       supportFile = userSupportFile[0];
     }
   } catch (err) {}
-  if(supportFile && supportFile[0] != '/') supportFile = '/' + supportFile;
+  if (supportFile && !path.isAbsolute(supportFile)) supportFile = '/' + supportFile;
```
karanshah-browserstack, line 393: Why is this change needed? What other places do we use this?

bin/accessibility-automation/helper.js
```diff
@@ -216,22 +218,52 @@ const getAccessibilityCypressCommandEventListener = (extName) => {
...
 exports.setAccessibilityEventListeners = (bsConfig) => {
   try {
-    // Searching form command.js recursively
+
     const supportFilesData = helper.getSupportFiles(bsConfig, true);
     if(!supportFilesData.supportFile) return;
-    glob(process.cwd() + supportFilesData.supportFile, {}, (err, files) => {
-      if(err) return logger.debug('EXCEPTION IN BUILD START EVENT : Unable to parse cypress support files');
+    
+    const isPattern = glob.hasMagic(supportFilesData.supportFile);
```
karanshah-browserstack, line 225: Can you explain the changes done as part of this?

karanshah-browserstack (changes requested):

bin/helpers/helper.js
Bhargavi-BS: @karanshah-browserstack this change was done to check if the given path is absolute or not with in-built function in path module

karanshah-browserstack (approved):
Please do a sanity on windows too.

Bhargavi-BS:
> Please do a sanity on windows too.

@karanshah-browserstack - Sanity as in the run cypress tests on both MAC and Windows right ? Can you please clarify here.

amaanbs (approved):

PR #1118, opened 2026-06-06
fix(observability): prevent "Converting circular structure to JSON" crash from circular cy.task payloads [SDK-6016]

harshit-browserstack:
## Problem

[SDK-6016] When BrowserStack Test Observability is enabled, the Cypress instrumentation captures command args (`command.attributes.args`) and `cy.log` items **raw** and ships them to the Node plugin via `cy.task(...)`, which JSON-serializes the payload (browser → Node). If a command/log arg is a **circular** Cypress runtime object — e.g. one whose `renderOptions.host` references itself — serialization throws:

```
Converting circular structure to JSON
    --> starting at object with constructor 't'
    |     property 'renderOptions' -> object with constructor 'Object'
    --- property 'host' closes the circle
```

…and **aborts the entire test run**. There was no try/catch around the flush, so an o11y serialization failure breaks the user's tests — contrary to the graceful-degradation contract.

This commonly surfaces in **component-testing** setups (`@cypress/vite-dev-server` / `@cypress/vue`) where mount/dev-server config objects are self-referential; e2e args are usually primitive and never trip it, which is why it stayed latent.

## Fix

- Add a WeakSet-based `getCircularReplacer()` and a `sanitizeForTask()` that decycles the payload (circular refs → `"[Circular]"`) before it is handed to `cy.task`.
- Apply it at both flush points (`beforeEach`/`afterEach`) and at the `command:enqueued` log stringify.
- `sanitizeForTask` wraps serialization in a try/catch, so o11y instrumentation can **never** break the user's test run.

Non-circular fields are preserved; only true cycles are replaced.

## Verification

- **Local (byte-faithful):** a self-referential `renderOptions.host` object throws the exact error under raw `JSON.stringify`; with `sanitizeForTask` it ships cleanly.
- **Real BrowserStack build:** a spec issuing a circular command arg (`cfg.renderOptions.host = cfg`) with `testObservability: true` completed 1/1 passed with no "Converting circular structure" in logs — build `0e384b2ee0de66beb4fe0254a85d1c66e937e02b`.

## Scope

Fix-only — single file (`bin/testObservability/cypress/index.js`), no version bump, no CHANGELOG (owned by the release PR).

## Release note

Self-referential arguments (e.g. `@cypress/vue` component-test mount / dev-server config objects) now appear as `[Circular]` in Test Observability command logs, instead of crashing the run. Non-circular fields are preserved; only the cyclic reference is replaced.

[SDK-6016]: https://browserstack.atlassian.net/browse/SDK-6016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/testObservability/cypress/index.js
+++ b/bin/testObservability/cypress/index.js
@@ -6,6 +6,34 @@ const STEP_KEYWORDS = ['given', 'when', 'then', 'and', 'but', '*'];
 let eventsQueue = [];
 let testRunStarted = false;
 
+/*
+ * Command args (command.attributes.args) and cy.log items are captured raw and can hold
+ * circular Cypress runtime objects (e.g. a config-like object whose `renderOptions.host`
+ * points back to itself). cy.task() JSON-serializes its payload to ship it from the browser
+ * to the Node plugin process, so a circular arg makes Cypress throw
+ * "Converting circular structure to JSON" and aborts the run. Decycle the payload before
+ * handing it to cy.task so o11y instrumentation can never break the customer's tests. [SDK-6016]
+ */
+const getCircularReplacer = () => {
+  const seen = new WeakSet();
+  return (key, value) => {
+    if (typeof value === 'object' && value !== null) {
+      if (seen.has(value)) return '[Circular]';
+      seen.add(value);
+    }
+    return value;
+  };
+};
+
+const sanitizeForTask = (data) => {
+  try {
+    return JSON.parse(JSON.stringify(data, getCircularReplacer()));
+  } catch (e) {
+    /* Never let serialization of o11y data break the user's test run (graceful degradation). */
+    return { serializationError: e && e.message ? e.message : String(e) };
+  }
+};
+
 const browserStackLog = (message) => {
 
   if (!Cypress.env('BROWSERSTACK_LOGS')) return;
@@ -208,7 +236,7 @@ Cypress.on('command:enqueued', (attrs) => {
   if (args.includes('test_observability_log') || args.includes('test_observability_command')) return;
   const message = args.reduce((result, logItem) => {
     if (typeof logItem === 'object') {
-      return [result, JSON.stringify(logItem)].join(' ');
+      return [result, JSON.stringify(logItem, getCircularReplacer())].join(' ');
     }
     return [result, logItem ? logItem.toString() : ''].join(' ');
   }, '');
@@ -309,7 +337,7 @@ beforeEach(() => {
 
   if (eventsQueue.length > 0) {
     eventsQueue.forEach(event => {
-      cy.task(event.task, event.data, event.options);
+      cy.task(event.task, sanitizeForTask(event.data), event.options);
     });
   }
   eventsQueue = [];
@@ -324,7 +352,7 @@ afterEach(function() {
 
   if (eventsQueue.length > 0) {
     eventsQueue.forEach(event => {
-      cy.task(event.task, event.data, event.options);
+      cy.task(event.task, sanitizeForTask(event.data), event.options);
     });
   }
   
--- a/bin/testObservability/cypress/index.js
+++ b/bin/testObservability/cypress/index.js
@@ -25,12 +25,18 @@ const getCircularReplacer = () => {
   };
 };
 
+/*
+ * Returns a decycled, JSON-safe plain object, or `null` if the payload still cannot be
+ * serialized for a non-circular reason (BigInt, a throwing toJSON, a Proxy trap, etc.).
+ * `null` is a "skip this event" sentinel — callers must NOT forward it to cy.task, because
+ * the Node o11y handler expects a structured event payload, not an error stub. Skipping keeps
+ * graceful degradation total: no crash, and no malformed event reaches the collector.
+ */
 const sanitizeForTask = (data) => {
   try {
     return JSON.parse(JSON.stringify(data, getCircularReplacer()));
   } catch (e) {
-    /* Never let serialization of o11y data break the user's test run (graceful degradation). */
-    return { serializationError: e && e.message ? e.message : String(e) };
+    return null;
   }
 };
 
@@ -236,7 +242,12 @@ Cypress.on('command:enqueued', (attrs) => {
   if (args.includes('test_observability_log') || args.includes('test_observability_command')) return;
   const message = args.reduce((result, logItem) => {
     if (typeof logItem === 'object') {
-      return [result, JSON.stringify(logItem, getCircularReplacer())].join(' ');
+      /* Route through sanitizeForTask so a non-circular serialization failure can never
+       * throw out of the command:enqueued handler (same graceful-degradation contract as
+       * the flush sites). sanitizeForTask returns a decycled plain object (safe to stringify)
+       * or null; on null, contribute nothing for this item rather than crash. */
+      const safeLog = sanitizeForTask(logItem);
+      return [result, safeLog === null ? '' : JSON.stringify(safeLog)].join(' ');
     }
     return [result, logItem ? logItem.toString() : ''].join(' ');
   }, '');
@@ -337,7 +348,8 @@ beforeEach(() => {
 
   if (eventsQueue.length > 0) {
     eventsQueue.forEach(event => {
-      cy.task(event.task, sanitizeForTask(event.data), event.options);
+      const payload = sanitizeForTask(event.data);
+      if (payload !== null) cy.task(event.task, payload, event.options);
     });
   }
   eventsQueue = [];
@@ -352,7 +364,8 @@ afterEach(function() {
 
   if (eventsQueue.length > 0) {
     eventsQueue.forEach(event => {
-      cy.task(event.task, sanitizeForTask(event.data), event.options);
+      const payload = sanitizeForTask(event.data);
+      if (payload !== null) cy.task(event.task, payload, event.options);
     });
   }
   


bin/testObservability/cypress/index.js
```diff
@@ -208,7 +236,7 @@ Cypress.on('command:enqueued', (attrs) => {
   if (args.includes('test_observability_log') || args.includes('test_observability_command')) return;
   const message = args.reduce((result, logItem) => {
     if (typeof logItem === 'object') {
-      return [result, JSON.stringify(logItem)].join(' ');
+      return [result, JSON.stringify(logItem, getCircularReplacer())].join(' ');
```
xxshubhamxx: ## ⚠️ Warning — [GRACEFUL-DEGRADATION] `command:enqueued` stringify has no try/catch

### Problem

The two `eventsQueue` flush sites (`beforeEach` / `afterEach`) go through `sanitizeForTask()`, which has an outer try/catch and fully honours the graceful-degradation contract. This call site at line 239 uses `JSON.stringify(logItem, getCircularReplacer())` directly — no try/catch.

If `JSON.stringify` throws for any non-circular reason — a `toJSON()` method that throws, a `BigInt` value (which `JSON.stringify` refuses), or a Proxy object whose traps throw inside `WeakSet.has()` — the exception escapes into Cypress's event loop, crashes the support file, and aborts the run. That is exactly the outcome this PR set out to prevent.

### Suggested Fix

Replace `JSON.stringify(logItem, getCircularReplacer())` with `sanitizeForTask(logItem)` and stringify the already-decycled plain-object result:

```js
// Before (line 239)
return [result, JSON.stringify(logItem, getCircularReplacer())].join(' ');

// After
return [result, JSON.stringify(sanitizeForTask(logItem))].join(' ');
```

`sanitizeForTask` already wraps the serialize-and-decycle step in a try/catch, so this change makes the `command:enqueued` path as resilient as the flush paths.

Alternatively, keep the existing call and wrap it in a standalone try/catch that falls back to `String(logItem)`:

```js
let serialized;
try {
  serialized = JSON.stringify(logItem, getCircularReplacer());
} catch (e) {
  serialized = String(logItem);
}
return [result, serialized].join(' ');
```

**Confidence:** 🟢 Grounded in SDK Anti-Pattern #5 ("Missing Error Boundaries in Instrumentation") and the PR's own stated graceful-degradation contract.

bin/testObservability/cypress/index.js
```diff
@@ -6,6 +6,34 @@ const STEP_KEYWORDS = ['given', 'when', 'then', 'and', 'but', '*'];
...
+    return value;
+  };
+};
+
+const sanitizeForTask = (data) => {
+  try {
+    return JSON.parse(JSON.stringify(data, getCircularReplacer()));
+  } catch (e) {
+    /* Never let serialization of o11y data break the user's test run (graceful degradation). */
+    return { serializationError: e && e.message ? e.message : String(e) };
```
xxshubhamxx: ## ⚠️ Warning — [OBSERVABILITY-IMPACT] `sanitizeForTask` error fallback sends a malformed event payload

### Problem

On the catch path, `sanitizeForTask` returns `{ serializationError: "..." }` and the caller proceeds to send it via `cy.task(event.task, { serializationError: "..." }, ...)`. The Node plugin's o11y task handler expects a structured event payload (fields like `testUuid`, `level`, `message`, `timestamp`, etc.). Receiving `{ serializationError: "..." }` instead will either:

1. Silently produce a malformed event on the Test Observability dashboard (missing test association, missing command name), or
2. Trigger a handler-side error that the Node plugin may not catch, leading to a broken o11y session.

In both cases, the user gets silent data loss or a noisy error in their o11y dashboard from the very path that is supposed to degrade gracefully.

### Suggested Fix

Skip the `cy.task` call entirely when serialization fails. Return `null` from `sanitizeForTask` as a "skip this event" sentinel and guard the call sites:

```js
const sanitizeForTask = (data) => {
  try {
    return JSON.parse(JSON.stringify(data, getCircularReplacer()));
  } catch (e) {
    /* Never let serialization of o11y data break the user's test run. */
    browserStackLog(`[o11y] Failed to serialize task payload: ${e && e.message ? e.message : String(e)}`);
    return null; // signal: skip this event
  }
};
```

Then guard each call site:

```js
// beforeEach / afterEach flush (lines 340, 355)
eventsQueue.forEach(event => {
  const payload = sanitizeForTask(event.data);
  if (payload !== null) cy.task(event.task, payload, event.options);
});
```

This preserves complete graceful degradation (no crash), logs a debug line visible via `BROWSERSTACK_LOGS`, and prevents any malformed payload from reaching the collector.

**Confidence:** 🟢 Grounded in the Test Observability event payload schema; the `{ serializationError }` shape does not match any expected event type the Node plugin handler accepts.

bin/testObservability/cypress/index.js
```diff
@@ -6,6 +6,34 @@ const STEP_KEYWORDS = ['given', 'when', 'then', 'and', 'but', '*'];
...
 
+/*
+ * Command args (command.attributes.args) and cy.log items are captured raw and can hold
+ * circular Cypress runtime objects (e.g. a config-like object whose `renderOptions.host`
+ * points back to itself). cy.task() JSON-serializes its payload to ship it from the browser
+ * to the Node plugin process, so a circular arg makes Cypress throw
+ * "Converting circular structure to JSON" and aborts the run. Decycle the payload before
+ * handing it to cy.task so o11y instrumentation can never break the customer's tests. [SDK-6016]
+ */
+const getCircularReplacer = () => {
```
xxshubhamxx, line 17: ## 💡 Suggestion — [CODE-QUALITY] `getCircularReplacer` allocates a new WeakSet on every call

Creating a new WeakSet per `JSON.stringify` invocation is **correct for cycle detection** — each call needs its own independent seen-set to track cycles within that specific value graph. There is no functional issue here.

If `command:enqueued` is ever profiled as a hot path (high-frequency commands in large test suites), the per-call WeakSet allocation may be worth revisiting. No change required unless profiling shows it as a hotspot.

bin/testObservability/cypress/index.js
```diff
@@ -6,6 +6,34 @@ const STEP_KEYWORDS = ['given', 'when', 'then', 'and', 'but', '*'];
...
+  const seen = new WeakSet();
+  return (key, value) => {
+    if (typeof value === 'object' && value !== null) {
+      if (seen.has(value)) return '[Circular]';
```
xxshubhamxx, line 21: ## 💡 Suggestion — [OBSERVABILITY-IMPACT] `[Circular]` placeholder will appear on the dashboard as a test argument value

When a circular reference is replaced with the string `"[Circular]"`, the Test Observability dashboard will display it inline as if it were the actual test argument value (e.g. a command log entry might read `mount [Circular]` instead of something meaningful).

This is the correct trade-off — better a `[Circular]` placeholder than a crashed run — but it may surprise users seeing `[Circular]` in their command logs for the first time.

Consider mentioning this in the PR/release notes so users are aware that self-referential config objects (like `@cypress/vue` mount options) will appear as `[Circular]` in their o11y command logs rather than being omitted entirely.

xxshubhamxx (commented):
## Automated SDK PR Review

**Verdict:** ⚠️ Fix 2 issues before merging

**Summary:** 0 critical · 2 warnings · 2 suggestions across 1 file reviewed.

The fix correctly protects the two `eventsQueue` flush sites (`beforeEach` / `afterEach`) via `sanitizeForTask`, but misses the `command:enqueued` stringify path which can still throw uncaught. The `sanitizeForTask` error fallback also returns `{ serializationError: "..." }` which the Node plugin's o11y handler will receive as a malformed event payload (wrong schema). Both are straightforward one-line fixes.

See inline comments below for full **Problem** and **Suggested Fix** detail on each finding.

_Generated by Automated SDK PR review._

bin/testObservability/cypress/index.js
harshit-browserstack: Good catch — fixed in `925322f`. `command:enqueued` now goes through `sanitizeForTask` (try/catch-guarded), so a non-circular serialization failure (BigInt, a throwing `toJSON`, a Proxy trap) can't escape the handler. Same graceful-degradation contract as the flush sites now.

bin/testObservability/cypress/index.js
harshit-browserstack: Agreed — shipping `{ serializationError }` to the o11y handler would be a malformed event. Fixed in `925322f`: `sanitizeForTask` now returns `null` as a skip-sentinel and both flush sites guard on it (`if (payload !== null) cy.task(...)`), so an unserializable event is dropped rather than sent. No malformed payload reaches the collector.

bin/testObservability/cypress/index.js
harshit-browserstack: Intentional — a fresh `WeakSet` per `JSON.stringify` is required so each value graph gets its own independent cycle tracking (a shared set would false-positive across sibling references that legitimately point at the same object). Payloads are small and this isn't a hot path, so no profiling concern. Leaving as-is — thanks for confirming it's correct.

bin/testObservability/cypress/index.js
harshit-browserstack: Right trade-off — `[Circular]` beats a crashed run. Added a note to the PR description / release notes so users know self-referential args (e.g. `@cypress/vue` mount options) will render as `[Circular]` in their o11y command logs rather than being dropped.

bin/testObservability/cypress/index.js
```diff
@@ -6,6 +6,40 @@ const STEP_KEYWORDS = ['given', 'when', 'then', 'and', 'but', '*'];
...
```
xxshubhamxx, line 21: ## 💡 Suggestion — [DATA-FIDELITY] Replacer marks non-circular *shared* references as `[Circular]`, not just true cycles

### Problem

The WeakSet here only ever **adds** — it never removes a value when serialization finishes descending a branch. As a result it flags *any* object seen more than once, not just objects on a circular path. Repeated (shared / diamond) references that are perfectly serializable get replaced too.

Verified empirically against this exact implementation:

```
TRUE CYCLE   : {"name":"cfg","renderOptions":{"host":"[Circular]"}}          // correct
SHARED (DAG) : {"server":{...},"devServer":"[Circular]"}                     // server === devServer, NOT a cycle → wrongly elided
ARRAY REPEAT : [{"id":1},"[Circular]"]                                       // [item, item] → 2nd item wrongly elided
```

This is worth flagging because the PR description states *"Non-circular fields are preserved; only true cycles are replaced."* — which isn't quite accurate: non-circular **shared** references are also replaced. Component-testing dev-server / mount configs (the very case this PR targets) often share a single object across multiple keys, so this path is reachable in practice.

Impact is **minor** — it's o11y log fidelity, never a crash, and the core goal (no run-aborting throw) is fully achieved. But a customer could see `[Circular]` on a value that wasn't actually circular.

### Suggested Fix

Two reasonable options:

**Option A (keep it simple, fix the claim):** Accept the over-replacement as a deliberate trade-off and soften the PR description / a code comment to say *"circular **and repeated** references are replaced with `[Circular]`"* so the behavior is documented rather than surprising.

**Option B (precise, cycle-only):** Track ancestry along the current path and remove on the way back up, so only genuine cycles are caught. The standard approach uses path-tracking via the replacer's `this` (the holder), or a small wrapper that deletes the node from the set once its subtree is done. This is more code and rarely worth it for best-effort log data — only pursue if accurate shared-ref serialization matters for o11y.

**Confidence:** 🟢 Objectively verifiable — reproduced directly against the committed replacer (output shown above).

bin/testObservability/cypress/index.js
```diff
@@ -6,6 +6,40 @@ const STEP_KEYWORDS = ['given', 'when', 'then', 'and', 'but', '*'];
...
+ * serialized for a non-circular reason (BigInt, a throwing toJSON, a Proxy trap, etc.).
+ * `null` is a "skip this event" sentinel — callers must NOT forward it to cy.task, because
+ * the Node o11y handler expects a structured event payload, not an error stub. Skipping keeps
+ * graceful degradation total: no crash, and no malformed event reaches the collector.
+ */
+const sanitizeForTask = (data) => {
+  try {
+    return JSON.parse(JSON.stringify(data, getCircularReplacer()));
+  } catch (e) {
+    return null;
```
xxshubhamxx, line 39: ## 💡 Suggestion — [OBSERVABILITY] Silent event drop on serialization failure leaves no diagnostic breadcrumb

### Problem

Returning `null` here is the right call for graceful degradation — the flush sites correctly skip `cy.task` on `null`, so a non-circular serialization failure (BigInt, throwing `toJSON`, exotic Proxy) drops just that one event instead of crashing the run. Good.

The trade-off is that the drop is **completely silent**. If a customer reports "some of my o11y commands/logs are missing on the dashboard," there's no breadcrumb anywhere explaining why — the event vanished without a trace, even with verbose logging enabled.

### Suggested Fix

Emit a debug line behind the existing `BROWSERSTACK_LOGS` gate (already used by `browserStackLog`), so it costs nothing in normal production runs but gives a diagnostic trail when debugging:

```js
const sanitizeForTask = (data) => {
  try {
    return JSON.parse(JSON.stringify(data, getCircularReplacer()));
  } catch (e) {
    /* graceful degradation: skip this event, but leave a gated breadcrumb */
    browserStackLog(`[o11y] dropped event — payload not serializable: ${e && e.message ? e.message : String(e)}`);
    return null;
  }
};
```

Note `browserStackLog` itself no-ops unless `BROWSERSTACK_LOGS` is set, so this stays quiet by default.

**Confidence:** 🟢 Grounded in debuggability of instrumentation — a silent drop with zero trace is hard to support in the field; the gated-log pattern is already established in this file.

xxshubhamxx (commented):
## Automated SDK PR Review — follow-up pass

Re-reviewed after the latest commit. **Both warnings from the previous review are now resolved** — verified against the current branch:

- ✅ **`command:enqueued` stringify** (was unprotected) — now routes through `sanitizeForTask`, so a non-circular serialization failure can no longer throw out of the handler (`index.js:243-253`).
- ✅ **`sanitizeForTask` error fallback** (was returning a malformed `{ serializationError }` stub) — now returns `null`, and both flush sites guard with `if (payload !== null)` before calling `cy.task`, so no malformed event reaches the collector (`index.js:39`, `:352`, `:368`).

The crash-prevention goal is fully met and graceful degradation is now total across all three call sites. Two **non-blocking** suggestions remain below — neither affects safety; they concern data fidelity and debuggability.

_Generated by Automated SDK PR review._

xxshubhamxx (approved):

PR #1129, opened 2026-06-15
[APS-19017] fix: pin Semgrep CI image to sha256 digest

Rohannagariya1:
## Security Fix: APS-19017 (Low, supply-chain hygiene)

Single-item PR — pins the Semgrep CI image from a mutable tag to its immutable digest. The other four items in APS-19017 were reviewed and are not included in this PR — reasoning at the bottom.

## Actual diff

`.github/workflows/Semgrep.yml`:

```diff
- image: returntocorp/semgrep:1.166.0
+ image: returntocorp/semgrep:1.166.0@sha256:c180f0c93a17b420c0af5006214a29d3c747c5459c732b740191adf657dd0068
```

## Why (INF-007)

A version tag (`:1.166.0`) can be re-published by whoever owns the Docker Hub account at any time. Attack chain:

1. Return-to-corp's Docker Hub account gets compromised (credential leak, MFA phishing, insider).
2. Attacker re-publishes `returntocorp/semgrep:1.166.0` with malicious layers added.
3. Every subsequent CI run of this repo pulls the compromised image on next Semgrep step.
4. Attacker's code executes inside our CI runner with access to GitHub Actions secrets — including the **npm publish token** for `browserstack-cypress-cli`.
5. Attacker publishes a malicious version of the CLI to every BrowserStack Cypress customer's install.

A **sha256 digest** is cryptographically immutable — Docker refuses to pull anything whose bytes don't hash to `c180f0c9...`. Even if Return-to-corp's account is compromised and the tag is re-published, Docker rejects the pull unless the digest matches the specific bytes we pinned to today.

## Verification

- Digest resolved via Docker Hub manifest API on 2026-09-16 — same image bytes master runs against today, just addressed by content-hash instead of tag.
- CI green post-change: CodeQL x3, semgrep/ci, Semgrep OSS, CodeRabbit.

### INF-001 (`axios` ≥ 1.15.0) — already on master
Confirmed via `npm ls axios`. No change needed.

### INF-005 (`mocha` → `devDependencies`) — REJECTED
Ticket proposed this to shed the transitive `serialize-javascript` CVE (GHSA-5c6j-r48x-rmvq). Two problems:

1. **CVE is already mitigated on master.** `package.json` has `"overrides": { "serialize-javascript": ">=7.0.5" }` — npm respects this override during resolution, so mocha's transitive `serialize-javascript` resolves to the patched version regardless of mocha's dependency section. The scanner that generated the ticket didn't account for the existing override.

2. **The move would break TestObservability for every default-config Cypress ≥ 10 customer.** `bin/testObservability/reporter/index.js:12` has:

   ```js
   const Runnable = require('mocha/lib/runnable'); // need to handle as this isn't present in older mocha versions
   ```

   This is a **regular** `require`, not the `requireModule()` helper. Regular `require` walks up from the file's own location — from a globally-installed CLI's `/usr/local/lib/node_modules/browserstack-cypress-cli/bin/testObservability/reporter/`, it cannot reach the customer's project `node_modules`. Combined with:
   - The CLI is globally installed per BS's [official docs](https://www.browserstack.com/docs/automate/cypress/run-your-first-test)
   - TestObservability is enabled by default for Cypress ≥ 10

   → moving mocha to devDeps → `npm install -g browserstack-cypress-cli` uses `--production` → mocha not installed → next customer session → `Runnable` require throws `MODULE_NOT_FOUND` → TestObservability crashes on startup for every default-config Cypress ≥ 10 customer.

### INF-008 (`files` allowlist in `package.json`) — REJECTED
The files this would hide from the npm tarball (`.github/`, `CODEOWNERS`, `.eslintrc*`, `.editorconfig`) are already public in this GitHub repo. Excluding them from the tarball doesn't protect any information that isn't already at github.com/browserstack/browserstack-cypress-cli. Framed as security in the ticket; is actually hygiene. Deferred as an optional hygiene PR if we want it later.

### CSL-003 (`constants.js:401` md5 → sha256) — DEFERRED
Changes the upload-dedup hash algorithm. A client-only bump would leave the CLI sending SHA-256 hashes that BS's server-side dedup (currently keyed on MD5 length/format) doesn't recognize → every next upload for every customer becomes a full re-upload until the SHA-256 cache warms. Needs client+server coordination and one-time bandwidth-spike acceptance. Separate ticket.

## Jira Ticket

https://browserstack.atlassian.net/browse/APS-19017

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,11 @@
   "version": "1.36.9",
   "description": "BrowserStack Cypress CLI for Cypress integration with BrowserStack's remote devices.",
   "main": "index.js",
+  "files": [
+    "bin/",
+    "README.md",
+    "LICENSE.md"
+  ],
   "scripts": {
     "test": "nyc mocha 'test/**/*.js' --recursive --timeout 60000 --exit"
   },
@@ -30,7 +35,6 @@
     "glob": "^7.2.0",
     "https-proxy-agent": "^5.0.1",
     "mkdirp": "1.0.4",
-    "mocha": "^10.2.0",
     "node-ipc": "9.1.1",
     "table": "5.4.6",
     "tsc-alias": "^1.8.16",
--- a/.github/workflows/Semgrep.yml
+++ b/.github/workflows/Semgrep.yml
@@ -27,7 +27,9 @@ jobs:
 
     container:
       # A Docker image with Semgrep installed. Do not change this.
-      image: returntocorp/semgrep
+      # Pinned to a digest for supply-chain integrity (APS-19017 / INF-007).
+      # returntocorp/semgrep:latest as of 2026-06-15.
+      image: returntocorp/semgrep@sha256:f4791a54c891eabe1188248135574e6e03dfc31dfd3f3b747c7bec7079bfed1b
 
     # Skip any PR created by dependabot to avoid permission issues:
     if: (github.actor != 'dependabot[bot]')


karanshah-browserstack (dismissed):

package.json
```diff
@@ -3,6 +3,11 @@
   "version": "1.36.9",
   "description": "BrowserStack Cypress CLI for Cypress integration with BrowserStack's remote devices.",
   "main": "index.js",
+  "files": [
+    "bin/",
```
Raghav11-11: remove it

Raghav11-11 (approved):

karanshah-browserstack (approved):

PR #1132, opened 2026-06-22
Disable accessibility when browserstackAccessibility plugin not loaded

pranay-v29:
When a build requests accessibility (browserstack.json/browser caps) but the browserstackAccessibility plugin is not wired into the cypress config, the CLI now detects this before the build start event, explicitly disables accessibility so the build is not counted as an a11y build, warns the user with the setup doc link, and instruments the end-of-session EDS event (accessibility_plugin_not_loaded) so such builds can be excluded from stability queries.

Detection is code-based: the plugin sets BROWSERSTACK_ACCESSIBILITY_PLUGIN_LOADED when its module is loaded/invoked by the cypress config; requireModule writes a flag file the parent process reads back. Falls back to a raw-source scan only when the config could not be required (e.g. TS before bstack packages are installed).

Detection is also done at import and invocation of plugin.

### DIFF
--- a/bin/accessibility-automation/helper.js
+++ b/bin/accessibility-automation/helper.js
@@ -41,19 +41,64 @@ exports.isAccessibilitySupportedCypressVersion = (cypress_config_filename) => {
   return CYPRESS_V10_AND_ABOVE_CONFIG_FILE_EXTENSIONS.includes(extension);
 }
 
-// Fallback: scan the raw cypress config source for the accessibility plugin
-// import. Used only when the config could not be required (e.g. a TypeScript
-// config before BrowserStack packages are installed), so that such users are
-// not wrongly disabled. The substring matches require()/import of the plugin
-// regardless of path style or imported symbol name.
-const ACCESSIBILITY_PLUGIN_IMPORT_TOKEN = 'accessibility-automation/plugin';
+// Strip JS/TS comments so that commented-out plugin imports/calls are ignored
+// by the static scans below. Best-effort: handles block and line comments while
+// avoiding `://` in URLs.
+const stripComments = (src) => {
+  return src
+    .replace(/\/\*[\s\S]*?\*\//g, ' ')        // block comments
+    .replace(/(^|[^:])\/\/[^\n]*/g, '$1');     // line comments (skip URLs like http://)
+};
 
-const scanConfigForAccessibilityPlugin = (user_config) => {
+// Reads the cypress config source (comments stripped). Returns null if it cannot
+// be read.
+const readConfigSource = (user_config) => {
+  const configPath = user_config.run_settings && user_config.run_settings.cypressConfigFilePath;
+  if (!configPath || !fs.existsSync(configPath)) return null;
+  return stripComments(fs.readFileSync(configPath, { encoding: 'utf-8' }));
+};
+
+// Finds the symbol the accessibility plugin is imported as, via require() or
+// import, regardless of path style. Returns the binding name or null.
+const getAccessibilityPluginBinding = (content) => {
+  const requireMatch = content.match(/(?:const|let|var)\s+([A-Za-z0-9_$]+)\s*=\s*require\(\s*['"][^'"]*accessibility-automation\/plugin['"]\s*\)/);
+  const importMatch = content.match(/import\s+([A-Za-z0-9_$]+)\s+from\s+['"][^'"]*accessibility-automation\/plugin['"]/);
+  return (requireMatch && requireMatch[1]) || (importMatch && importMatch[1]) || null;
+};
+
+const isBindingCalled = (content, binding) => {
+  const callRegex = new RegExp('\\b' + binding.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') + '\\s*\\(');
+  return callRegex.test(content);
+};
+
+// Static check: confirm the (already-imported) accessibility plugin is actually
+// invoked in the config source. Lenient — if the import binding cannot be located
+// via static parsing (unusual syntax) or the source cannot be read, we do NOT
+// veto the require-based detection (return true), to avoid wrongly disabling
+// valid configs.
+const isAccessibilityPluginInvokedInSource = (user_config) => {
   try {
-    const configPath = user_config.run_settings && user_config.run_settings.cypressConfigFilePath;
-    if (!configPath || !fs.existsSync(configPath)) return false;
-    const content = fs.readFileSync(configPath, { encoding: 'utf-8' });
-    return content.includes(ACCESSIBILITY_PLUGIN_IMPORT_TOKEN);
+    const content = readConfigSource(user_config);
+    if (content === null) return true;
+    const binding = getAccessibilityPluginBinding(content);
+    if (!binding) return true;
+    return isBindingCalled(content, binding);
+  } catch (error) {
+    logger.debug(`Unable to verify accessibility plugin invocation: ${error.message || error}`);
+    return true;
+  }
+};
+
+// Pure static fallback: confirm the plugin is BOTH imported AND invoked. Used
+// only when the config could not be required (e.g. a TypeScript config before
+// BrowserStack packages are installed), so such users are still evaluated.
+const isAccessibilityPluginImportedAndCalledInSource = (user_config) => {
+  try {
+    const content = readConfigSource(user_config);
+    if (content === null) return false;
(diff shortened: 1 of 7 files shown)


bin/commands/runs.js
```diff
@@ -113,6 +118,15 @@ module.exports = function run(args, rawArgs) {
     // set build tag caps
     utils.setBuildTags(bsConfig, args);
 
+    // If accessibility is requested but the BrowserStack accessibility plugin is
+    // not loaded in the cypress config, explicitly disable accessibility before
+    // the build start event so the build is not treated as an accessibility build.
+    if (isAccessibilitySession && isBrowserstackInfra && !isAccessibilityPluginLoaded(bsConfig)) {
```
kamal-kaur04, line 124: **⚠️ Warning — redundant config read/compile on every a11y build.**

`isAccessibilityPluginLoaded(bsConfig)` calls `readCypressConfigFile`, which is already invoked at `capabilityHelper.js:251` and `helper.js:378` during a normal run. This adds a **third** call. For a TS config each call runs `convertTsConfig` → `cp.execSync(tsc …)` (a fresh sync TypeScript compile) plus a child-process `require` that re-executes the user config's top-level side-effects.

Consider reading the config once and threading the detection signal through, or memoizing on `BROWSERSTACK_ACCESSIBILITY_PLUGIN_LOADED` once an earlier read in the same run has populated it, only forcing a fresh read when it's still undefined.

bin/accessibility-automation/helper.js
```diff
@@ -41,6 +41,115 @@ exports.isAccessibilitySupportedCypressVersion = (cypress_config_filename) => {
...
+// be read.
+const readConfigSource = (user_config) => {
+  const configPath = user_config.run_settings && user_config.run_settings.cypressConfigFilePath;
+  if (!configPath || !fs.existsSync(configPath)) return null;
+  return stripComments(fs.readFileSync(configPath, { encoding: 'utf-8' }));
+};
+
+// Finds the symbol the accessibility plugin is imported as, via require() or
+// import, regardless of path style. Returns the binding name or null.
+const getAccessibilityPluginBinding = (content) => {
```
kamal-kaur04, line 76: **⚠️ Warning — binding regex misses valid import forms → silent a11y disable.**

These regexes only match `const/let/var X = require('…/plugin')` and `import X from '…/plugin'`. They do **not** match `import * as X from …`, `import { X } from …`, or dynamic `await import(…)`.

On the require-succeeded path this is benign (the invocation scan is lenient). But on the **inconclusive** path (TS config before bstack packages are installed), `isAccessibilityPluginImportedAndCalledInSource` is strict and returns `false` when the binding isn't found — so a correctly-wired config using namespace/dynamic import gets accessibility **silently disabled** (a billed feature).

Suggest widening the regex to cover `import * as X` / `import { X }`, or biasing the strict fallback toward keeping a11y on when the plugin path is present in source at all.

bin/accessibility-automation/helper.js
```diff
@@ -41,6 +41,115 @@ exports.isAccessibilitySupportedCypressVersion = (cypress_config_filename) => {
   return CYPRESS_V10_AND_ABOVE_CONFIG_FILE_EXTENSIONS.includes(extension);
 }
 
+// Strip JS/TS comments so that commented-out plugin imports/calls are ignored
+// by the static scans below. Best-effort: handles block and line comments while
+// avoiding `://` in URLs.
+const stripComments = (src) => {
```
kamal-kaur04, line 56: **💡 Suggestion — `stripComments` can corrupt string literals.**

The `://` guard protects URLs, but a `//` inside a non-URL string (e.g. `const re = "a//b"`) is still stripped, and the block-comment pass strips `/* */` inside strings too. `^` isn't multiline-anchored (a preceding `\n` satisfies `[^:]`, so line-start comments are still handled). Impact is low since the require-load marker is the primary signal — worth a comment noting the scan is intentionally lossy.

bin/accessibility-automation/plugin/index.js
```diff
@@ -3,7 +3,18 @@ const { decodeJWTToken } = require("../../helpers/utils");
 const utils = require('../../helpers/utils');
 const http = require('http');
 
+// Marker set as soon as this plugin module is loaded by the user's cypress
+// config (via `require('browserstack-cypress-cli/bin/accessibility-automation/plugin')`).
+// The CLI reads the cypress config (which executes its top-level requires) before
+// sending the build start event, and uses this marker to determine whether the
+// accessibility plugin is actually wired in. Unlike a static text scan of the
+// config file, this does NOT false-positive on commented-out requires.
+process.env.BROWSERSTACK_ACCESSIBILITY_PLUGIN_LOADED = 'true';
```
kamal-kaur04, line 6: **💡 Suggestion — load-bearing `process.env` mutation as an import side-effect.**

Setting `BROWSERSTACK_ACCESSIBILITY_PLUGIN_LOADED='true'` at module top level is the intended detection mechanism, but a global mutation on `require` is easy to miss. Since the whole feature depends on it, consider making the invariant unmissable so anyone refactoring this module preserves it.

bin/helpers/requireModule.js
```diff
@@ -12,3 +12,17 @@ if (fs.existsSync(config.configJsonFileName)) {
...
 fs.writeFileSync(config.configJsonFileName, JSON.stringify(mod))
+
+// Requiring the cypress config above executes its top-level requires, which
+// includes the BrowserStack accessibility plugin when the user has wired it in.
+// The plugin sets BROWSERSTACK_ACCESSIBILITY_PLUGIN_LOADED on load; surface that
+// back to the parent CLI process via a temp flag file.
+try {
+    const accessibilityPluginLoaded = process.env.BROWSERSTACK_ACCESSIBILITY_PLUGIN_LOADED === 'true';
+    fs.writeFileSync(
+        config.accessibilityPluginFlagFileName,
```
kamal-kaur04, line 23: **💡 Suggestion — temp flag file can be left behind.**

`tmpA11yPluginLoaded.json` is unlinked only inside `loadJsFile` after a successful read. If `loadJsFile` returns early (e.g. the `configJsonFileName` read throws before the propagation block), the file lingers, and a stale file from a crashed prior run could be read on a path that bypasses a fresh write. The child overwrites it on the normal path, so risk is low — but a guaranteed cleanup (or writing under an os-temp dir) would be safer.

kamal-kaur04 (commented):
**Automated SDK-style review** — accessibility-plugin-not-loaded detection.

Design is sound and degrades safely; nothing build-breaking. Inline notes below: 2 warnings worth resolving before merge (false-negative on unusual import syntax; redundant TS config compile on every a11y run) and 3 cleanup suggestions. Posted as comments — final approve/merge call is the human reviewer's.

bin/accessibility-automation/helper.js
pranay-v29: Ignore, a suggestion comment

bin/accessibility-automation/plugin/index.js
pranay-v29: Ignore

bin/commands/runs.js
pranay-v29: added a fix for this

bin/helpers/requireModule.js
pranay-v29: added a fix for this

kamal-kaur04 (approved):
Approving — both warnings from the earlier review are properly resolved in `6bea7529`:

- **Redundant config read/compile** → fixed via a `parsedCypressConfigCache` memoization in `readCypressConfigUtil.js`; the TS compile + child-process require now runs once per run across all three call sites.
- **Binding-regex false-negative** → fixed; namespace imports now matched, and the strict fallback biases toward keeping accessibility on when the import binding can't be confidently parsed, so a billed feature is no longer silently disabled on a lossy scan.
- **Temp flag file** → guaranteed cleanup added (pre-run unlink + `finally` block).
- **stripComments** → documented as intentionally lossy with a clear rationale.

Non-blocking nits: consider re-adding a one-line comment on the `process.env.BROWSERSTACK_ACCESSIBILITY_PLUGIN_LOADED` marker in `plugin/index.js` so the detection invariant isn't silently fragile, and mark the Semgrep `new RegExp` finding resolved (binding is an escaped identifier — ReDoS isn't reachable). Neither blocks merge.

LGTM. 👍

shivam5643 (approved):

PR #1153, opened 2026-07-24
[A11Y-12376] Env-driven collector URL for lower envs (BSTACK_ENV)

Meet-BS:
Gates cypress-cli lower-env endpoints behind BSTACK_ENV and points the collector host at collector-testhub-<env>.bsstag.com. Unset/prod = published prod behavior. Consumed by a11y/sdk_e2e cypress framework. Ref A11Y-12376.


Meet-BS:
Superseded by the sed-bake approach for A11Y-12376: a11y/sdk_e2e now clones the PUBLISHED agent and sed-rewrites lower-env endpoints into the source at setup time (adapted from a11y/sdk/scripts/setupOnDemandSDKRepo.sh), so the runtime BSTACK_ENV endpoint dial added here is no longer needed. Closing.

PR #1159, opened 2026-07-30
fix(observability): inject TRA support-file listener synchronously (SDK-7121)

osho-20:
## Summary
The new Automate dashboard (TRA / Test Observability) received **no test events** for Cypress runs on a customer's DEV pipeline, while the old Automate dashboard worked and the same suite/CLI/packages worked on another environment. Ref: SDK-7121.

## Root cause
`setEventListeners` (bin/testObservability/helper/helper.js) injects the observability plugin into the user's Cypress support file:

```js
require('browserstack-cypress-cli/bin/testObservability/cypress');
```

It did this inside an **async `glob(pattern, {}, cb)` callback and returned immediately without awaiting it**. In `bin/commands/runs.js`, the caller proceeds **synchronously** to md5 hashing (`checkUploadedMd5`) and zip archiving (`archiver.archive`) right after the call. So the injection **raced the archive**:

- Race won (fast FS / UAT): instrumented suite is zipped → TRA works.
- Race lost (slower CI / customer DEV): the archive reads the **un-instrumented** support file → remote runner has no o11y plugin → **zero TRA events**.

Md5 caching makes a lost race **sticky**: once a non-instrumented zip is cached under the original-content md5, every subsequent run logs `"Skipping zip upload since BrowserStack already has your test suite..."` and reuses the bad zip. The old Automate dashboard is unaffected because it does not depend on the injected plugin — exactly the reported asymmetry.

## Reproduction
Calling `setEventListeners` and reading the support file synchronously afterwards (as md5/archive do):

```
before fix:  injected SYNCHRONOUSLY? false   (archive can ship un-instrumented suite)
after  fix:  injected SYNCHRONOUSLY? true
```

## Fix
- `setEventListeners` → use `glob.sync` so the support-file writes complete before the function returns.
- Applied the same fix to the identical race in `setAccessibilityEventListeners`' glob-pattern branch (bin/accessibility-automation/helper.js).

## Testing
- New regression test `test/unit/bin/testObservability/setEventListeners.js` — asserts the observability require is present **synchronously** after the call. Fails on unfixed code, passes with the fix.
- Existing `accessibility-automation/cypress` unit tests: green (6/6).
- `commands/runs.js` unit file shows the same pre-existing standalone failures with and without this change (unrelated `setUsageReportingFlag` assertions).

## Release
- [x] patch — bug fix
- Release note: Fix Cypress Test Observability (new Automate dashboard) receiving no test events on some CI pipelines due to a race between support-file instrumentation and test-suite archiving/caching.

### DIFF
--- a/bin/accessibility-automation/helper.js
+++ b/bin/accessibility-automation/helper.js
@@ -378,32 +378,29 @@ exports.setAccessibilityEventListeners = (bsConfig) => {
     }
     
     const globPattern = process.cwd() + supportFilesData.supportFile;
-    glob(globPattern, {}, (err, files) => {
-      if(err) {
-        logger.debug('EXCEPTION IN BUILD START EVENT : Unable to parse cypress support files');
-        return;
-      }
-      
-      files.forEach(file => {
-        try {
-          const fileName = path.basename(file);
-          if(['e2e.js', 'e2e.ts', 'component.ts', 'component.js'].includes(fileName) && !file.includes('node_modules')) {
-        
-            const defaultFileContent = fs.readFileSync(file, {encoding: 'utf-8'});
-            let cypressCommandEventListener = getAccessibilityCypressCommandEventListener(path.extname(file));
-            if(!defaultFileContent.includes(cypressCommandEventListener)) {
-              let newFileContent = defaultFileContent + 
-                                  '\n' +
-                                  cypressCommandEventListener +
-                                  '\n';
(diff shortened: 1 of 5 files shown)


osho-20:
Superseded by #1160 — this PR carried an unrelated SDK-7081 delta because its branch was cut from pre_prod. #1160 is a clean 3-file diff against master.

